PR #77, opened 2020-08-21
WIP: Add support for SQLite in tests

srtfisher:
Adds a `db.php` for unit tests to allow tests to be run against SQLite. This allows for some performance improvements since SQLite is faster than MySQL. But the real results will come if MNTL-157 is possible, and we can switch to SQLite in-memory.

Until then, will leave this as a WIP since the benefits of switching to SQLite using a file store isn't as strong as SQLite in-memory.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -423,16 +423,16 @@
         },
         {
             "name": "league/commonmark",
-            "version": "1.5.1",
+            "version": "1.5.4",
             "source": {
                 "type": "git",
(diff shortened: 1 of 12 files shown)


anubisthejackle:
I had a thought about this but haven't tested it yet. It might be possible to create a separate process accessible like an API that maintains the sqlite memory thread, and have the tests database call out to that. The question would then be if that would be fast enough to be worth it.

PR #97, opened 2020-10-14
MNTL-165 generator generators

dustinpaluch:
Add a generator to generate generators.
`wp mantle make:generator <name> <type>`

Generates the generator class, and creates a basic stub template for that specific generator.

There is a bug with passing a `type` that has backslashes. https://alleyinteractive.atlassian.net/browse/MNTL-185

Doesn't handle capitalization very well, i.e. ideally, `wp mantle make:generator foo` would give you `class Foo_...` and `protected $name = 'make:foo';`.

### DIFF
--- a/src/mantle/framework/console/generators/stubs/generator.stub
+++ b/src/mantle/framework/console/generators/stubs/generator.stub
@@ -27,7 +27,6 @@ class {{ class }}_Make_Command extends Generator_Command {
 	 */
 	protected $description = 'Generate a {{ class }}.';
 
-
 	/**
 	 * Command synopsis.
 	 *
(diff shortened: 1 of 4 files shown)


src/mantle/framework/console/generators/stubs/generator.stub
```diff
@@ -0,0 +1,70 @@
...
+
+	/**
+	 * Command Description.
+	 *
+	 * @var string
+	 */
+	protected $description = 'Generate a {{ class }}.';
+
+
+	/**
```
srtfisher, line 31:
```suggestion

	/**
```

src/mantle/framework/console/generators/stubs/generator.stub
```diff
@@ -0,0 +1,70 @@
...
+	 * The type of class being generated.
+	 *
+	 * @var string
+	 */
+	protected $type = '{{ type }}';
+
+
+	/**
```
srtfisher, line 53:
```suggestion

	/**
```

srtfisher (approved):
sweet!

PR #158, opened 2021-02-20
PHP 8

### LINKED ISSUE #134: Test and Support PHP 8

srtfisher:
Closes #134 

Mostly deals with changes to the Reflection class and the `is_callable()` not respecting non-static functions (that aren't being called statically!)


renatonascalves (approved):
🗡️

PR #170, opened 2021-07-10
Asset Manager Provider

srtfisher:
Integrate with https://github.com/alleyinteractive/wp-asset-manager to provide a means to manage assets easily in WordPress.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,6 +10,7 @@
         }
     ],
     "require": {
+        "alleyinteractive/wp-asset-manager": "dev-composer",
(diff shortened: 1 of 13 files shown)


composer.json
```diff
@@ -10,6 +10,7 @@
         }
     ],
     "require": {
+        "alleyinteractive/wp-asset-manager": "dev-composer",
```
srtfisher, line 13: Need to update after merging in PR from https://github.com/alleyinteractive/wp-asset-manager

PR #204, opened 2021-11-16
Remove need for /bin/install-wp-tests.sh in unit tests

### LINKED ISSUE #200: Remove/improve bin/install-wp-tests.sh
Remove or improve the need to run `bash bin/install-wp-tests.sh` before running tests in a CI environment. Locally the plugin/theme will already be a WordPress installation naturally. When running a CI test, the project itself will be cloned out without the parent WordPress installation. We then need to `rsync` the project into a test folder and install WordPress. Maybe this can be improved so that we can run `composer run test` and everything else simply happens in the background.

srtfisher:
Removes the need for CI scripts and non-WordPress rooted projects from having to `rsync` their package to a WordPress installation. The main use case is a plugin can now run `composer run phpunit` and simplify have their plugin tested without having to setup any external WordPress installation.

Fixes #200


src/mantle/testing/wordpress-bootstrap.php
```diff
@@ -31,13 +33,56 @@
...
+	if ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) {
+		echo 'WordPress installation not found, installing in temporary directory: ' . WP_TESTS_INSTALL_PATH . PHP_EOL;
+
+		$dir = get_framework_directory();
+
+		// Download the latest installation command from GitHub and install WordPress.
+		$cmd = sprintf(
+			'WP_CORE_DIR=%s curl -s %s | bash -s %s %s %s %s %s %s',
+			WP_TESTS_INSTALL_PATH,
+			'https://raw.githubusercontent.com/alleyinteractive/mantle-framework/main/bin/install-wp-tests.sh',
```
mboynes, line 60: Because of `main`, this isn't a permalink. If we at some point move the file, or remove it, it would immediately break every mantle installation's integration of this code. On the other hand, having a dynamic reference carries benefits too, e.g. we can improve it or fix a bug and every codebase using it instantly benefits. Overall I think the cons of a dynamic reference outweigh the pros, so I'd recommend going that route. Two thoughts/ideas:

1. Instead of `main`, set it to a tag. That would be more future-proof -- but not entirely so. e.g. what if we renamed the repo?
2. Consider removing the shell script from this repo entirely and moving it to another one. It could even be, and likely should be, its own repo. Maybe there would even be value in pulling this file (or some of it) and the shell script into a separate composer package? Regardless, if that shell script had its own home, it would be safer to reference without thinking about it moving.

mboynes (approved):
❤️  love this update. pitched an idea for the bootstrapping process and the shell script. it's a blocker because I don't think this PR should ship as-is, but I'm still approving the PR so you can merge it once you pick a path on that. happy to help brainstorm some other ideas for that, too.

src/mantle/testing/wordpress-bootstrap.php
srtfisher: One idea would be to at least use the `HEAD` version instead of `main`: https://raw.githubusercontent.com/alleyinteractive/mantle-framework/HEAD/bin/install-wp-tests.sh Homebrew uses a standalone repository https://github.com/Homebrew/install .

#2 seems like a better option to decouple the two things

srtfisher: Ended up going with https://raw.githubusercontent.com/alleyinteractive/mantle-ci/HEAD/install-wp-tests.sh !

PR #209, opened 2021-12-28
Replacing events with native WordPress hook events

srtfisher:
- [x] Replaces the formerly Laravel-based events system with a system that integrates with WordPress hooks with some Mantle flair.
- [x] Removes some functionality that came with the Laravel based system (to be readded in the future, including queued event dispatching, and subscribers)
- [x] Automatic event discovery via listeners.

```php
class Example_Listener {
	/**
	 * Handle the event.
	 *
	 * @param Example_Event $event
	 * @return void
	 */
	public function handle( Example_Event $event ) {
		//
	}

	public function handle_other_event_at_20( Example_Event $event ) {
		//
	}

	public function on_wp_loaded_at_20() {
		// ...
	}
}
```


src/mantle/framework/events/class-events-manifest.php
```diff
@@ -0,0 +1,159 @@
...
+	 */
+	protected function write_manifest( array $manifest ) {
+		$dir        = dirname( $this->manifest_path );
+		$filesystem = new Filesystem();
+
+		// Ensure the cached folder exists.
+		if ( ! is_dir( $dir ) ) {
+			$filesystem->ensure_directory_exists( $dir );
+
+			// Create the folder if it doesn't exist.
```
mogmarsh, line 143:
```suggestion
			// Throw an exception if the folder doesn't exist.
```

src/mantle/framework/events/class-events-manifest.php
```diff
@@ -0,0 +1,159 @@
...
+	 * @param array $manifest Manifest to write.
+	 * @throws Application_Exception Thrown on error writing file.
+	 */
+	protected function write_manifest( array $manifest ) {
+		$dir        = dirname( $this->manifest_path );
+		$filesystem = new Filesystem();
+
+		// Ensure the cached folder exists.
+		if ( ! is_dir( $dir ) ) {
+			$filesystem->ensure_directory_exists( $dir );
```
mogmarsh, line 141:
```suggestion
			// Create the folder if it doesn't exist.
			$filesystem->ensure_directory_exists( $dir );
```

mogmarsh (approved):
wow.. I didn't 100% understand everything going on there, but everything looked pretty good. I did catch one misleading comment, but that was it. 👍

PR #214, opened 2021-12-29
Ensure the status is set once and returns properly for the rest api

srtfisher:


tests/testing/concerns/test-makes-http-requests.php
```diff
@@ -81,6 +96,11 @@ public function test_rest_api_route() {
 			->assertJsonPath( 'title.rendered', get_the_title( $post_id ) );
 	}
 
+	public function test_rest_api_route_error() {
+		$this->get( rest_url('/an/unknown/route' ) )
```
srtfisher, line 100: PHP fixes are incoming in another PR :)

mogmarsh (approved):
👍

PR #215, opened 2021-12-29
Allow cookies to be set when testing

srtfisher:
- Adding test support for cookies
- Removing server var reference


mogmarsh (approved):
🍪

PR #216, opened 2021-12-30
Add dd/dump commands to test response

srtfisher:
Add a method to dump/debug the response from the test response.

```php
class Example_Test {
	public function test_rest_api_route_error() {
		$this->get( rest_url( '/an/unknown/route' ) )
			->assertStatus( 404 )
			->dump_headers()
			->dd();
	}
}
```
Screen Shot 2021-12-30 at 2 29 05 PM


mslinnea (approved):
Nice! 🐴

PR #218, opened 2022-01-04
Additional tests for request tests

srtfisher:


tests/testing/concerns/test-makes-http-requests.php
```diff
@@ -104,29 +120,14 @@ public function test_rest_api_route_error() {
...
-		$_SERVER['__cookie_test'] = false;
-
-		$this->app['router']->get(
-			'/set-cookie',
-			function() {
-				$_SERVER['__cookie_test'] = request()->cookie( 'test-cookie', 'fallback-value' );
-
-				return 123;
+	public function test_multiple_requests() {
+		// Re-run all test methods on this class in a single pass.
```
mslinnea, line 124: This is an interesting test... 🤔

mslinnea (approved):
🐴

tests/testing/concerns/test-makes-http-requests.php
srtfisher: Yeah, I ran into some issues where the tests could be run in isolation but not one after another. Figured I might as well re-run them all in a single go!

PR #223, opened 2022-01-07
Simplify the automagic loading of classes

srtfisher:
Uses the `Mantle\Support\Traits\Loads_Classes` instead of the `Finder`


mogmarsh (approved):
👍

PR #236, opened 2022-02-11
Using alleyinteractive/wordpress-autoloader for autoloading

srtfisher:
- https://github.com/alleyinteractive/wordpress-autoloader
- https://github.com/alleyinteractive/composer-wordpress-autoloader

- [x] Track the 0.1.1 release of alleyinteractive/composer-wordpress-autoloader after https://github.com/alleyinteractive/composer-wordpress-autoloader/pull/2 is merged.

### DIFF
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -19,6 +19,11 @@ namespace Mantle;
  * @return \Closure Function for spl_autoload_register().
  */
 function generate_wp_autoloader( string $namespace, string $root_path ): callable {
+	trigger_error( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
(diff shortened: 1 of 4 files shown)


src/autoload.php
```diff
@@ -10,66 +10,17 @@
 /**
  * Generate an autoloader for the WordPress file naming conventions.
  *
+ * @deprecated Use https://github.com/alleyinteractive/wordpress-autoloader or
```
daronspence, line 13: Would we want to add a `trigger_error` as well? From some light googling, that seems to be good thing to include. Not sure if that would affect any production projects.

daronspence (approved):
🤘 one comment but not blocking

src/autoload.php
srtfisher: Great idea!

PR #242, opened 2022-04-01
Mono-repo support

srtfisher:
- Moves kernel into the framework namespace.
- Adds `composer.json` for all packages.
- Moves helpers into `Mantle\Support`

Includes the already reviewed https://github.com/alleyinteractive/mantle-framework/pull/247


monorepo-builder.php
```diff
@@ -0,0 +1,72 @@
...
+		],
+	);
+
+	// Append specific items to the composer.json.
+	$parameters->set(
+		Option::DATA_TO_APPEND,
+		[
+			ComposerJsonSection::REQUIRE     => [
+				'alleyinteractive/composer-wordpress-autoloader' => '^0.2',
+				'php' => '^7.4|^8.0',
```
anubisthejackle, line 51:
```suggestion
				'alleyinteractive/composer-wordpress-autoloader' => '^0.2',
				'php'                                            => '^7.4|^8.0',
```

src/mantle/container/class-container.php
```diff
@@ -662,9 +662,9 @@ protected function get_contextual_concrete( $abstract ) {
 			return $binding;
 		}
 
-			// Next we need to see if a contextual binding might be bound under an alias of the
-			// given abstract type. So, we will need to check if any aliases exist with this
-			// type and then spin through them and check for contextual bindings on these.
+		// Next we need to see if a contextual binding might be bound under an alias of the
+		// given abstract type. So, we will need to check if any aliases exist with this
+		// type and then spin through them and check for contextual bindings on these.
```
anubisthejackle, line 667:
```suggestion
		/**
		 * Next we need to see if a contextual binding might be bound under an alias of the
		 * given abstract type. 
		 * 
		 * To do that, we will need to check if any aliases exist with this type and then 
		 * spin through them and check each for contextual bindings as well.
		 */
```

anubisthejackle (approved):
I didn't see anything that shouted wrong to me. Couple style suggestions, but no blockers.
🎸

PR #248, opened 2022-04-02
Add testing library database connect helpers

srtfisher:


src/mantle/framework/console/class-test-config-install-command.php
```diff
@@ -68,7 +68,13 @@ public function handle( array $args, array $assoc_args = [] ) {
...
 		}
 
-		$this->log( __( 'Configuration copied! Update it to reference your proper database credentials.', 'mantle' ) );
+		$this->log(
+			sprintf(
+				$path,
+				/* translators: 1: config file path */
+				__( 'Configuration copied to %s. Update it to reference your proper database credentials.', 'mantle' ),
+			),
+		);
```
anubisthejackle, line 77: I think these parameters are reversed, I'm assuming `$path` is meant to satisfy the `%s`?

anubisthejackle (approved):
The `sprintf` needs to be fixed before this is merged, but other then that it looks good to me. :guitar:

PR #249, opened 2022-04-06
Ensure all shell script variables are valid arguments

srtfisher:
Ensure that setting `WP_DB_PASSWORD=""` as an environmental variable doesn't break the installation. Currently that has to be set to `WP_DB_PASSWORD="''"` to work properly.


src/mantle/testing/class-utils.php
```diff
@@ -232,4 +232,19 @@ public static function env( string $variable, $default ) {
 
 		return false === $value ? $default : $value;
 	}
+
+	/**
+	 * Ensure an environment variable is is a valid string that can be passed to
```
dustinpaluch, line 237: `is is`

dustinpaluch (approved):

PR #269, opened 2022-05-18
Create new Testkit cases: Integration and Unit Test

anubisthejackle:
This PR creates two new Testkit Test Case classes: `Integration_Test_Case` and `Unit_Test_Case`.

# Integration_Test_Case
This test case executes `\Mantle\Testing\install` on `setUpBeforeClass` once, bringing WordPress into the global scope. To allow us to continue to perform actions after install, the trait `Installs_WordPress` looks for an optional globally scoped method `mantle_after_wordpress_install` which, if found, will be used as the callback to `\Mantle\Testing\install`.

# Unit_Test_Case
This test case bypass Mantle entirely, extending directly from `PHPUnit\Framework\TestCase` and forcing `PreserveGlobalState` to false, and `RunClassInSeparateProcess` to true. This allows us to isolate our unit test's global scope from our Integration tests, which allows us to run them in tandem, without concerns of getting fatal errors around `PHP Fatal error: Cannot redeclare function`.

# Other changes
To make the above work, there were some changes to `\Mantle\Testing\Test_Case`:
* `get_test_case_traits` was converted to a static method.
* `setUpBeforeClass` now looks for the existance of a method `trait_name_set_up_before_class` for each loaded trait, and calls it statically, immediately prior to `parent::setUpBeforeClass`.

### DIFF
--- a/src/mantle/testkit/concerns/trait-installs-wordpress.php
+++ b/src/mantle/testkit/concerns/trait-installs-wordpress.php
@@ -1,0 +1,34 @@
+<?php
+/**
+ * Installs_WordPress trait file
+ *
+ * @package Mantle
+ */
+
+namespace Mantle\Testkit\Concerns;
+
+/**
+ * Concern for handling WordPress installation outside of the bootstrap.
+ */
+trait Installs_WordPress {
+	/**
+	 * Manages installing WordPress.
+	 *
+	 * This is useful for installing WordPress in integration tests, while leaving the bootstrap
+	 * free of this overhead during unit tests. To perform actions after WordPress is installed, you
+	 * simply need to define a function `mantle_after_wordpress_install` in your bootstrap file.
+	 */
+	public static function installs_wordpress_set_up_before_class(): void {
+		static $installed = false;
+
+		if( true === $installed ) {
+			return;
+		}
+
+		$callback = function_exists( 'mantle_after_wordpress_install' ) ? 'mantle_after_wordpress_install' : null;
+
+		\Mantle\Testing\install( $callback );
+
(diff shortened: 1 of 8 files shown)


src/mantle/testkit/concerns/trait-installs-wordpress.php
```diff
@@ -0,0 +1,34 @@
...
+	public static function installs_wordpress_set_up_before_class(): void {
+		static $installed = false;
+
+		if ( true === $installed ) {
+			return;
+		}
+
+		$callback = function_exists( 'mantle_after_wordpress_install' ) ? 'mantle_after_wordpress_install' : null;
+
+		\Mantle\Testing\install( $callback );
```
srtfisher, line 30: I wonder if this should be a filter instead of a pre-defined function? /shrug

srtfisher (approved):
looks great! if you get around to it, documenting this in https://github.com/alleyinteractive/mantle-docs/blob/main/testing/testkit.md would be wonderful

src/mantle/testkit/concerns/trait-installs-wordpress.php
anubisthejackle: I thought about that, too, but I don't think at this point we'd have access to filters, unless there is a way to do that built into Mantle. Another option could be to look for the existence of a class method, but that would require extending `Integration_Test_Case` which pulls it away from how `Unit_Test_Case` works, which seemed not ideal.

srtfisher: eh, not a blocker at all. i think it's good as is!

PR #270, opened 2022-05-19
Fix Faker deprecation warnings about accessing methods as parameters

anubisthejackle:
During some testing it became apparent that in certain configurations, Faker was throwing a deprecation warning which was causing some tests to fail.

This PR updates Mantle's use of Faker to avoid those deprecation warnings.

* Specific Faker line in question: https://github.com/FakerPHP/Faker/blob/acef8f29bfd88bb046a5fe2a08693977aa9501bd/src/Faker/Generator.php#L950


srtfisher (approved):
🏆

PR #273, opened 2022-05-20
Allow attributes to register hooks

srtfisher:
```php
use Mantle\Support\Attributes\Action;

class My_Provider extends Service_Provider {
	#[Action('my-hook', 20)]
	public function handle_custom_hook() {
		// Do some cool stuff!
	}
}
```


mogmarsh (approved):
one question, otherwise 👍

PR #274, opened 2022-05-23
Ensure WP_MULTISITE can be passed properly

srtfisher:
* Only set the `MULTISITE` constant for non-install requests. During the test lifecycle, the `wordpress-bootstrap.php` file loads `install-wordpress.php` via shell. `install-wordpress.php` then loads `wordpress-bootstrap.php` again which had set `MULTISITE` to true. With `MULTISITE` set, `ms-settings.php` was then loaded which caused a fatal error during installation. This constant should not be set during the internal installation of WordPress. This would be one difference Mantle has from core.
* WordPress was not being installed properly but was silently failing. A fix was put in place to check for a proper installation.
* Updated the message to reflect passing WP_MULTISITE=1 or defining WP_TESTS_MULTISITE=1 as a constant.

### DIFF
--- a/src/mantle/testing/wordpress-bootstrap.php
+++ b/src/mantle/testing/wordpress-bootstrap.php
@@ -126,13 +126,11 @@ if ( ! $installing_wp && '1' !== getenv( 'WP_TESTS_SKIP_INSTALL' ) ) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


mogmarsh (approved):
👍

PR #277, opened 2022-05-27
Define Post Model Terms Fluently

srtfisher:
Extends #276
Docs https://github.com/alleyinteractive/mantle-docs/pull/21/files

Allows post models to add/set/delete/update terms using a fluent format.

The `Post` model support interacting with terms through relationships or through the model directly. The model supports multiple methods to make setting terms on a post simple:

```php
$category = Category::whereName( 'Example Category' )->first();

// Save the category to a post.
$post = new Post( [ 'title' => 'Example Post' ] );

// Also supports an array of IDs or WP_Term objects.
$post->terms->category = [ $category ];

$post->save();

// Read the tags from a post.
$post->terms->post_tag // Term[]
```

Terms can also be set when creating a post (specifying the taxonomy is optional):

```php
$post = new Post( [
	'title' => 'Example Title',
	'terms' => [ $category ],
] );

$post = new Post( [
	'title' => 'Example Title',
	'terms' => [
		'category' => [ $category ],
		'post_tag' => [ $tag ],
	],
] );
```

### DIFF
--- a/src/mantle/database/model/term/trait-model-term.php
+++ b/src/mantle/database/model/term/trait-model-term.php
@@ -13,6 +13,7 @@ use Mantle\Support\Arr;
 use WP_Term;
 
 use function Mantle\Support\Helpers\collect;
+use function Mantle\Support\Helpers\get_term_object;
 
 /**
  * Interface for interfacing with a model's terms.
@@ -76,8 +77,17 @@ trait Model_Term {
 	 * Store queued model terms.
 	 */
 	protected function store_queued_terms() {
-		foreach ( $this->queued_terms as $taxonomy => $values ) {
-			$this->set_terms( $values, $taxonomy );
+		if ( empty( $this->queued_terms ) ) {
+			return;
+		}
+
+		// Determine if this is an array of terms instead of taxonomy => term pairs.
+		if ( Arr::is_assoc( $this->queued_terms ) ) {
+			foreach ( $this->queued_terms as $taxonomy => $values ) {
+				$this->set_terms( $values, $taxonomy );
+			}
+		} else {
+			$this->set_terms( $this->queued_terms );
 		}
 
 		$this->queued_terms = [];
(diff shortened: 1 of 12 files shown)


src/mantle/testing/concerns/trait-assertions.php
```diff
@@ -368,4 +371,82 @@ public function assertUserDoesNotExists( array $arguments ) {
...
+		}
+
+		$term = $this->get_term_from_argument( $term );
+
+		if ( $term ) {
+			PHPUnit::assertTrue( \has_term( $term->term_id, $term->taxonomy, $post ) );
+		} else {
+			PHPUnit::fail( 'Term not found to assert against' );
+		}
+	}
```
anubisthejackle, line 418: We can avoid the if/else here, and be more explicit, by checking that `$term` is actually a \WP_Term.

```suggestion
	public function assertPostHasTerm( $post, $term ) {
		if ( $post instanceof Post ) {
			$post = $post->id();
		}

		$term = $this->get_term_from_argument( $term );

		PHPUnit::assertInstanceOf( \WP_Term::class, $term, 'Term not found to assert against' );
		PHPUnit::assertTrue( \has_term( $term->term_id, $term->taxonomy, $post ) );
	}
```

src/mantle/testing/concerns/trait-assertions.php
```diff
@@ -368,4 +371,82 @@ public function assertUserDoesNotExists( array $arguments ) {
...
+		}
+
+		$term = $this->get_term_from_argument( $term );
+
+		if ( $term ) {
+			PHPUnit::assertFalse( \has_term( $term->term_id, $term->taxonomy, $post ) );
+		} else {
+			PHPUnit::fail( 'Term not found to assert against' );
+		}
+	}
```
anubisthejackle, line 441: I think the logic is off on this one a bit. We're verifying that a term doesn't exist, so we shouldn't fail if `$term` is empty.

```suggestion
	public function assertPostNotHasTerm( $post, $term ) {
		if ( $post instanceof Post ) {
			$post = $post->id();
		}

		$term = $this->get_term_from_argument( $term );
		
		PHPUnit::assertNotInstanceOf( \WP_Term::class, $term, 'Term found to assert against' );
		PHPUnit::assertFalse( \has_term( $term->term_id, $term->taxonomy, $post ) );
	}
```

src/mantle/database/model/term/trait-model-term.php
```diff
@@ -0,0 +1,226 @@
...
+	 * Remove terms from a post.
+	 *
+	 * @param mixed  $terms Accepts an array of or a single instance of terms.
+	 * @param string $taxonomy Taxonomy name, optional.
+	 * @return static
+	 *
+	 * @throws Model_Exception Thrown if the $taxonomy cannot be inferred from $terms.
+	 */
+	public function remove_terms( $terms, string $taxonomy = null ) {
+		$terms = collect( is_array( $terms ) ? $terms : [ $terms ] )
```
anubisthejackle, line 193:
```suggestion
		$terms = collect( Arr::wrap( $terms ) )
```

src/mantle/testing/concerns/trait-assertions.php
srtfisher: `get_term_from_argument()` has a return declaration of `?WP_Term` so we can always assume it'll be a `WP_Term` or null.

src/mantle/testing/concerns/trait-assertions.php
```diff
@@ -368,4 +371,77 @@ public function assertUserDoesNotExists( array $arguments ) {
...
+		if ( $post instanceof Post ) {
+			$post = $post->id();
+		}
+
+		$term = $this->get_term_from_argument( $term );
+
+		if ( $term ) {
+			PHPUnit::assertFalse( \has_term( $term->term_id, $term->taxonomy, $post ) );
+		}
+	}
```
anubisthejackle, line 436: I know this has already been merged, but I wanted to call out that without `PHPUnit::assertNotInstanceOf` we no longer have an assertion if `$term` is null. That was the intent behind that line in my previous suggestion.

PR #285, opened 2022-06-24
Adding a With_Faker trait

srtfisher:


src/mantle/testing/concerns/trait-with-faker.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/**
+ * This file contains the Admin_Screen trait
```
mogmarsh, line 3:
```suggestion
 * This file contains the With_Faker trait
```

mogmarsh (approved):
one comment fix. 👍

PR #286, opened 2022-06-24
Assert Json Structure and other testing improvements

srtfisher:
- Renaming to Test_Interacts_With_External_Requests to clarify the intent of the test
- Moving all JSON tests to `Assertable_Json_String` -- a broken out class for asserting against JSON. More tests will be added in the future to it!
- Adding assertJsonStructure method
- Wiring up the response to take a JsonSerializable straight


mogmarsh (approved):
looks good 👍

PR #301, opened 2022-07-27
Adding Concurrent Http Client Request Support

srtfisher:
Adding support for concurrent HTTP requests with the Http Client package using core `wp_remote_request()` through the use of https://github.com/alleyinteractive/wp-concurrent-remote-requests/.


mogmarsh (approved):
👍

PR #302, opened 2022-07-27
Adding a Installation_Manager to facilitate installation

srtfisher:
Adds an installation manager to allow folks to fluently install the Mantle Testing Framework and define before/after hooks.

## Previously

```php
Mantle\Testing\install(
	function() {
		tests_add_filter( 'muplugins_loaded', '_manually_load_environment' );
	}
);
```

## New

```php
Mantle\Testing\manager()
	->before( ... )
	->after( ... )
	->loaded( ... )
	->install();
```

I've also combined the logic of `mantle_after_wordpress_install` to this manager (it's a singleton) but the logic of `mantle_after_wordpress_install` will still continue to exist.


src/mantle/testing/class-installation-manager.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @param int      $priority Priority.
+	 * @return static
+	 */
+	public function on( string $hook, ?callable $callback, int $priority = 10 ) {
+		if ( is_callable( $callback ) ) {
+			tests_add_filter( $hook, $callback, $priority );
+		}
+
+		return $this;
+	}
```
anubisthejackle, line 91:
```suggestion
	/**
	 * Define a callback for a specific WordPress hook.
	 *
	 * @param string   $hook Hook name.
	 * @param callable $callback Callback to invoke.
	 * @param int      $priority Priority.
	 * @param int      $accepted_args Number of accepted arguments.
	 * @return static
	 */
	public function on( string $hook, ?callable $callback, int $priority = 10, int $accepted_args = 1 ) {
		if ( is_callable( $callback ) ) {
			tests_add_filter( $hook, $callback, $priority, $accepted_args );
		}

		return $this;
	}
```

anubisthejackle (approved):
Looks good. I like the fluent API. Added a suggestion for passing `$accepted_args` but that's all I can see.

PR #304, opened 2022-07-28
Adding Conditionable Method Chaining

srtfisher:
Allows for fluent method chaining:

```php
Posts::where('name', 'this-name')
  ->when( ..., function ($query) {
    $query->whereMeta( 'key' , 'value' );
  })
  ->get();
```


mogmarsh (approved):
👍

PR #305, opened 2022-07-28
Gutenberg Block Service Provider

anubisthejackle:
This PR adds support for Gutenberg Blocks, including:

* A Blocks service provider that handles autoloading blocks found in `/app/blocks`
* An abstract block class that all blocks extend from that handles enqueue front and backend assets, as well as loading the view for dynamic blocks.
* A block generator that generates:
    * A block class
    * An entry (index.js)
    * An edit.jsx
    * An attributes.json 
    * A block view blade template

## Generator Command
`wp mantle make:block <name> <block_namespace> <block_name> [--description] [--category] [--icon]`

All flags are optional, but some of that information is required (Description, Category). If those flags aren't used, a prompt will be provided during load that forces the user to enter data for those fields.


src/mantle/framework/blocks/class-discover-blocks.php
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	/**
+	 * Extract the class name from a given file path.
+	 *
+	 * @param SplFileInfo $file File.
+	 * @param string      $base_path Base path.
+	 * @return string
+	 */
+	protected static function class_from_file( SplFileInfo $file, string $base_path ): string {
```
srtfisher, line 82: `Mantle\Support\Traits\Loads_Classes` exists to do some of this

anubisthejackle: Dropping this in where I would expect to use it, it isn't returning the correct class name. I suspect there's something I'm missing.

src/mantle/framework/console/generators/class-block-make-command.php
```diff
@@ -0,0 +1,421 @@
...
+	 * @param string  $question The question to ask in the input.
+	 * @param ?string $default The optional default value for the response.
+	 * @return callable A callable to be used as a default value for the `flag` method.
+	 */
+	protected function require_input( string $question, ?string $default = null ): string {
+		$response = $this->input( $question );
+
+		/**
+		 * If we don't get a response, recurse until we do, unless we have a defined default value.
+			*/
```
srtfisher, line 414:
```suggestion
		 */
```

srtfisher (approved):
i'm 💯 loving this! I do have some thoughts outside of this PR on what is the best path forward for asset management in mantle. I had a PR stub #319 that would actually move towards a traditional alley-webpack structure vs mix (since mix is a bit of an outlier around here). i'd love to get your thoughts on that part as we head down this road and optionally kill/merge #319.

anubisthejackle:
@srtfisher That's a good question. I lean towards keeping Mix and growing our usage of it across the company for a couple of reasons.

1. Mantle is technically open source, and as such, it might make sense to keep Mix as it's gaining in popularity, with a lot of documentation available that we don't have to maintain.
2. Mix is also open source, so we benefit as a project, and as a company, by improvements made outside of Alley to the project if we continue to use it.

I'm sure there are more reasons, but those are my main two.

srtfisher:
@anubisthejackle agree with you -- unless things change and an Alley solution presents itself I'd love to offload it all to Mix.

PR #309, opened 2022-08-05
Ensure tests have a permalink structure by default

srtfisher:
- Ensure that tests can have a permalink structure by default.
- Ensure Mantle tests against single and multisite.


mogmarsh (approved):
👍

PR #310, opened 2022-08-05
Adding only_children() method to collections

srtfisher:
Adds a neat method to retrieve only specific child array keys in a collection.


src/mantle/support/class-collection.php
```diff
@@ -789,6 +789,28 @@ public function only( $keys ) {
 		return new static( Arr::only( $this->items, $keys ) );
 	}
 
+	/**
+	 * Get the items in an collection of arrays with filtered child keys.
+	 *
+	 * @param    mixed $keys
+	 * @return static
```
mogmarsh, line 796: spacing and label for `$keys`

mogmarsh (approved):
one little thing, otherwise 👍

PR #322, opened 2022-09-15
Use separate packages for Caper and filter side effects

dlh01:
Replaces these features with the new standalone packages that have been split from what used to be an internal library.

If this PR is accepted, the docs repo will need to be updated (which I can do), and something, I assume, will need to happen to https://github.com/mantle-framework/caper (which I can't do).

I'm having some trouble getting the tests to run locally, so I'll be leaning on the results from GitHub Actions to know whether this update has been made correctly.


srtfisher (approved):
ah! you read my mind. love it! I have a 0.6.0 release queued up that I'd love to include this in.

PR #323, opened 2022-09-17
Asset assertions and improvements

srtfisher:
- Add assertions for asset enqueuing.
- Make it easier to enqueue assets in the admin/editor/etc.

Part of #118

### DIFF
--- a/src/mantle/assets/class-asset.php
+++ b/src/mantle/assets/class-asset.php
@@ -307,7 +307,7 @@ class Asset {
 	/**
 	 * Tell the asset whether or not to load in the block editor
 	 *
-	 * @param bool $load True if this should load in the admin area of the site.
+	 * @param bool $load True if this should load in the block editor.
 	 * @return Asset
 	 */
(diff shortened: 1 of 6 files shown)


src/mantle/assets/class-asset.php
```diff
@@ -86,14 +86,21 @@ class Asset {
...
-	protected $frontend = true;
+	protected bool $frontend = true;
 
 	/**
 	 * Enqueue in the admin area.
 	 *
 	 * @var bool
 	 */
-	protected $admin = true;
+	protected bool $admin = false;
```
anubisthejackle, line 96: This changes previous behavior where assets were included front and back by default. If that's intended, it should be noted somewhere.

anubisthejackle (approved):
Aside from that note about BC change, looks good to go :guitar:

src/mantle/assets/class-asset.php
srtfisher: Reverted!

PR #325, opened 2022-09-23
Support for mix pulling in dependencies

srtfisher:
- Adds support for mix pulling in dependencies from @wordpress/dependency-extraction-webpack-plugin.
- Adds tests for Mix.

Supports #118


mogmarsh (approved):
👍

PR #335, opened 2022-10-20
Middleware for testing factories

srtfisher:
Allows for fluent creation of posts with terms/meta. Adds support for middleware when factory objects are being created.

```
static::factory()->post->with_terms( ... )->create();
```

### DIFF
--- a/src/mantle/testing/factory/class-post-factory.php
+++ b/src/mantle/testing/factory/class-post-factory.php
@@ -50,7 +50,7 @@ class Post_Factory extends Factory {
 	/**
 	 * Create a new factory instance to create posts with a set of terms.
 	 *
-	 * @param array<int, WP_Term|int|string>|WP_Term|int|string ...$terms Terms to assign to the post.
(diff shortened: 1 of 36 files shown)


src/mantle/testing/factory/class-post-factory.php
```diff
@@ -46,46 +47,83 @@ public function __construct( Generator $generator, string $post_type = 'post' )
 		$this->post_type = $post_type;
 	}
 
+	/**
+	 * Create a new factory instance to create posts with a set of terms.
+	 *
+	 * @param array<int, WP_Term|int|string>|WP_Term|int|string ...$terms Terms to assign to the post.
```
mogmarsh, line 53:
```suggestion
	 * @param array<int, WP_Term|int|string>|WP_Term|int|string> ...$terms Terms to assign to the post.
```
I think this is right? If not, ignore me

mogmarsh (approved):
👍

PR #337, opened 2022-10-31
Add support for an isolated console mode

srtfisher:
- Adds support for running console commands (esp `wp mantle package:discover` and the other discovery commands without WordPress installed.
- Refactors the console to rely entirely on the Symfony Console. `wp mantle` is registered with wp-cli which will subsequently launch the Symfony console. This is the same console that can be fired by running `bin/mantle`.


mogmarsh (approved):
I found just a couple small things, otherwise 👍 
There's a lot going on here.

PR #338, opened 2022-11-03
Adding support for testing commands

srtfisher:
Adds support for testing commands in unit tests. This is currently limited to within the framework only with plans to expand this to WP CLI tests.


mogmarsh (approved):
👍 looks good!

PR #341, opened 2022-11-15
Use the existing WP_CORE_DIR if one exists

srtfisher:
Fixes a conflict @kevinfodness noted where Broadway already has `WP_CORE_DIR` defined. If the variable is defined and the site isn't running in an existing WordPress installation it will use that variable's setting to store where WordPress should be installed for testing.


kevinfodness (approved):
🍣

PR #350, opened 2022-12-12
Factory Updates

srtfisher:
- Adds more user information to the user factory for `display_name`, `first_name`, and `last_name`.
- Adds `with_meta()` support to term and user factories.
- Adds a `with_posts()` to the term factory.


mogmarsh (approved):
👍

PR #362, opened 2023-01-10
Upgrading to voku/portable-ascii 2.0

srtfisher:


srtfisher:
Merging to fix a downstream error with a bad dependency.

mogmarsh (commented):
👍

PR #366, opened 2023-02-17
New Asset Loader

srtfisher:
- Remove Laravel Mix
- Switch to asset loader that is aligned with https://github.com/alleyinteractive/create-wordpress-plugin


mboynes (approved):
💯

PR #368, opened 2023-02-26
Fix get_facade_accessor signature

szepeviktor:
it was a fatal error
@srtfisher Please consider using PHPStan.
https://github.com/alleyinteractive/mantle-framework/blob/b0b041d9e811a733cec98d3c3760448948defe4f/src/mantle/facade/class-facade.php#L40


szepeviktor:
Thank you.

PR #369, opened 2023-02-26
Make wp_insert_post return WP_Error

szepeviktor:
as it is tested with `is_wp_error`
https://developer.wordpress.org/reference/functions/wp_insert_post/

Please consider using PHPStan.


srtfisher:
@szepeviktor thanks for the PR! PHPStan is a big priority for myself and the rest of Alley this year.

srtfisher (approved):

PR #373, opened 2023-03-30
Improve reporting of stray HTTP requests

srtfisher:
Screenshot 2023-03-30 at 1 13 33 PM

Improve the reporting of stray HTTP requests. Previously stray requests were logged and reported when they were made. This works well for isolated unit tests but when a request was made on a page (when the output buffer is used), the log fell on silent ears. We'll now alert on tear down that a stray request was made.


mslinnea (approved):
🐴

PR #378, opened 2023-03-31
Use the core shim in the main test case

srtfisher:


anubisthejackle (approved):
:guitar:

src/mantle/testing/concerns/trait-wordpress-state.php
```diff
@@ -18,7 +18,7 @@ trait WordPress_State {
 	/**
 	 * Cleans the global scope (e.g `$_GET` and `$_POST`).
 	 */
-	public static function clean_up_global_scope() {
+	public function clean_up_global_scope() {
```
renatonascalves, line 21: Should you change here to use `$this`?

https://github.com/alleyinteractive/mantle-framework/blob/4051a5f2891c44b6b5fc15a4f0b9b5fe1c8d9e8f/src/mantle/testing/class-test-case.php#L156

renatonascalves (approved):
🗡️

src/mantle/testing/concerns/trait-wordpress-state.php
anubisthejackle: +1 to this

srtfisher: We all know this makes better sense as a static method 😭

PR #380, opened 2023-04-10
Add a with_image() helper to create an attachment with a real image

srtfisher:
Introduces the following to make attachment generation with an actual image file easier:

```php
static::factory()->attachment->with_image()->create()
```

### DIFF
--- a/src/mantle/testing/factory/class-attachment-factory.php
+++ b/src/mantle/testing/factory/class-attachment-factory.php
@@ -119,7 +119,7 @@ class Attachment_Factory extends Post_Factory {
 	 *
 	 * @return int|\WP_Error The attachment ID on success. The value 0 or WP_Error on failure.
 	 */
-	public function create_upload_object( $file, $parent = 0 ) {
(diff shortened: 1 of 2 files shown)


src/mantle/testing/factory/class-attachment-factory.php
```diff
@@ -48,56 +51,87 @@ public function create( array $args = [] ): ?int {
...
+	 * Saves an attachment.
+	 *
+	 * @deprecated Use the `with_image()` method instead.
+	 *
+	 * @param string $file   The file name to create attachment object for.
+	 * @param int    $parent ID of the post to attach the file to.
+	 *
+	 * @return int|\WP_Error The attachment ID on success. The value 0 or WP_Error on failure.
+	 */
+	public function create_upload_object( $file, $parent = 0 ) {
```
renatonascalves, line 122:
```suggestion
	public function create_upload_object( $file, $parent = 0 ): int|\WP_Error {
```

renatonascalves (approved):
🗡️

PR #383, opened 2023-04-14
Handle some edge cases when running wp-cli on VIP

srtfisher:
Ensure that the `wp-cli` binary is properly stripped as well as `path` which confused the Symfony application.


attackant (commented):
🍦

PR #393, opened 2023-05-30
Add support for using SQLite in tests (opt-in)

srtfisher:
Depends on https://github.com/alleyinteractive/mantle-ci/pull/11

Adds opt-in support for using SQLite in tests in place of MySQL. Uses the `db.php` drop-in from https://github.com/aaemnnosttv/wp-sqlite-db

### DIFF
--- a/src/mantle/testing/class-installation-manager.php
+++ b/src/mantle/testing/class-installation-manager.php
@@ -120,6 +120,16 @@ class Installation_Manager {
 	}
 
 	/**
+	 * Alias for `theme()`.
+	 *
+	 * @param string $theme Theme name.
+	 * @return static
+	 */
+	public function with_theme( string $theme ) {
+		return $this->theme( $theme );
+	}
+
+	/**
 	 * Define the active plugins to be set after the installation is loaded.
 	 *
 	 * @param array<int, string> $plugins Plugin files.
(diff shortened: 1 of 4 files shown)


nlemoine:
Hi @srtfisher, 

Looks like SQLite is on [its way to core](https://make.wordpress.org/core/2022/09/12/lets-make-wordpress-officially-support-sqlite/). Why not using the [official implementation](https://github.com/WordPress/sqlite-database-integration)?

Amazing work on Mantle by the way!

anubisthejackle (approved):
🎸

anubisthejackle:
> Why not using the [official implementation](https://github.com/WordPress/sqlite-database-integration)?

I think the main reason I can see is that the official plugin isn't a simple drop-in `db.php` which is what we're using for testing purposes.

srtfisher:
@nlemoine I'd love to! That is actually handled over in our CI script that we use (https://github.com/alleyinteractive/mantle-ci/pull/11) and I can add a TODO to my list to review switching it over. I will second @anubisthejackle here and say that it was mighty simple to get a single `db.php` added versus a whole set of files from the plugin. Hopefully we can switch over to core's https://wordpress.org/plugins/sqlite-database-integration/ soon enough.

PR #394, opened 2023-06-02
Introduce application bootloader

srtfisher:
Refactor the application to be less reliant on the boilerplate code in https://github.com/alleyinteractive/mantle and work more independently.

## Problem

Mantle projects often require a good amount of boilerplate code to use them (see https://github.com/alleyinteractive/mantle). This boilerplate code is prohibiting Mantle from reaching its fullest potential. For the most part, nobody prefers having to port over _all_ of this code to make Mantle run. It's holding back the greatness from within the framework.

## Solution

Introduce the Bootloader. The Bootloader abstracts the setting up of the framework with the right kernel given the current context. Though mirroring https://github.com/laravel/laravel is wonderful (boot the console kernel in `artisan`, boot the web kernel in `public/index.php`, etc.), it doesn't translate well WordPress land.

We still want applications to have the flexibility to do whatever they want with Mantle. But we want to allow for that while making it **as simple as possible** to run it without all the boilerplate code.

Here's an example of what is now possible (previously this required a whole bunch of setup code):

```php
// Boot the application.
bootloader()->boot();

// And get going!

// Register a router.
Route::get( '/example-route/', function () {
	dispatch( function () {
		file_put_contents(ABSPATH . '/example.txt', time() );
	} );

	return 'aye';
} );

Route::get( '/another-example/', function() {
	dispatch( function () {
		file_put_contents(ABSPATH . '/example.txt', time() );
	} );

	return response()->json( [ 'aye' => 'aye' ] );
} );
```

There is something to be said about how to make this even more extendable and easier to use. But it is a start!

Projects like `alleyinteractive/mantle` are now a starter kit of what can be done but no longer prescribes all the setup code needed. All you need is the above code and you should be good-to-go.

https://github.com/alleyinteractive/mantle/compare/0.12.x

### DIFF
--- a/src/mantle/events/class-dispatcher.php
+++ b/src/mantle/events/class-dispatcher.php
@@ -116,7 +116,11 @@ class Dispatcher implements Dispatcher_Contract {
 	public function dispatch( $event, $payload = [ null ] ) {
 		[ $event, $payload ] = $this->parse_event_and_payload( $event, $payload );
 
-		return apply_filters( $event, ...$payload ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.DynamicHooknameFound
+		if ( function_exists( 'apply_filters' ) ) {
+			return apply_filters( $event, ...$payload ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.DynamicHooknameFound
+		}
+
+		return null;
 	}
 
 	/**
--- a/src/mantle/framework/class-bootloader.php
+++ b/src/mantle/framework/class-bootloader.php
@@ -1,0 +1,275 @@
+<?php
+/**
+ * Bootloader class file
+ *
+ * @package Mantle
+ */
+
+namespace Mantle\Framework;
+
+use Closure;
+use Mantle\Application\Application;
+use Mantle\Console\Command;
+use Mantle\Contracts;
+use Mantle\Contracts\Framework\Bootloader as Contract;
+use Mantle\Http\Request;
+use Mantle\Support\Str;
+use Mantle\Support\Traits\Conditionable;
+
+/**
+ * Boot Manager
+ *
+ * Used to instantiate the application and load the framework given the current
+ * context. Removes the need for boilerplate code to be included in projects
+ * (ala laravel/laravel) but still allows for the flexibility to do so if they
+ * so choose.
+ */
+class Bootloader implements Contract {
+	use Conditionable;
+
+	/**
+	 * Current instance of the manager.
+	 *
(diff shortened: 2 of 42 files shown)


src/mantle/events/class-dispatcher.php
```diff
@@ -116,7 +116,11 @@ protected function resolve_subscriber( $subscriber ) {
 	public function dispatch( $event, $payload = [ null ] ) {
 		[ $event, $payload ] = $this->parse_event_and_payload( $event, $payload );
 
-		return apply_filters( $event, ...$payload ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.DynamicHooknameFound
+		if ( function_exists( 'apply_filters' ) ) {
+			return apply_filters( $event, ...$payload ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.DynamicHooknameFound
```
attackant, line 120: If apply filters doesn't exist, do we want to return these values anyway?

src/mantle/framework/class-bootloader.php
```diff
@@ -0,0 +1,275 @@
...
+
+				exit( (int) $status );
+			},
+			[
+				'shortdesc' => __( 'Mantle Framework Command Line Interface', 'mantle' ),
+			]
+		);
+	}
+
+	/**
```
attackant, line 265: Do we anticipate other contexts where we want to boot like web sockets or using a wp-now type setup or something else? Do we want to make this more extensible?

src/mantle/query-monitor/class-query-monitor-service-provider.php
```diff
@@ -68,7 +63,7 @@ public function fire_query_monitor_dispatches(): ?string {
 
 		foreach ( $this->query_monitor_dispatches as $callback ) {
 			// Remove the dispatcher from the 'shutdown' hook.
-			remove_action_validated( 'shutdown', $callback, 0 );
+			remove_action( 'shutdown', $callback, 0 );
 			$callback();
```
attackant, line 67: Curious what this was "remove_action_validated"

attackant (approved):
👍🏼

src/mantle/framework/class-bootloader.php
srtfisher: `wp-now` would fall into a HTTP request but sockets would be something to think about

src/mantle/events/class-dispatcher.php
srtfisher: Mostly a WIP to get my `bin/console` command working.

PR #396, opened 2023-06-09
Refresh application routing and add tests

srtfisher:
- Add `without_middleware()` support.
- Bring REST API routing into parity with the rest of the routing logic.
- Allow `prefix()` and `name()` to work when prefixing the name/prefix of the route.
- Fix HEAD requests.
- Add tests for URL generation.
- Adjust URL generation to make it generate a proper route URL for a named route.


src/mantle/http/routing/class-router.php
```diff
@@ -324,7 +356,20 @@ function( Request $request ) use ( $route ) {
...
 		// Ensure the response is valid since the middleware can modify it after it is run through Route.
-		return Route::ensure_response( $response );
+		return static::to_response( $request, $response );
+	}
+
+	/**
+	 * Prepare a response for sending.
+	 *
+	 * @param Request $request
+	 * @param mixed   $response
```
renatonascalves, line 366: Missing a proper description.

src/mantle/http/routing/class-router.php
```diff
@@ -492,33 +564,35 @@ public function substitute_implicit_bindings( Request $request ) {
...
+	 * @param string                $namespace Namespace for the REST API route.
+	 * @param callable|string       $callback  Callback that will be invoked to register
+	 *                                         routes OR a string route.
+	 * @param callable|array|string $args      Callback for the route if $callback is a
+	 *                                         string route OR arguments to pass to
+	 *                                         the register_rest_route() call. Not used if $callback
+	 *                                         is a closure.
 	 */
-	public function rest_api( string $namespace, $route, $args = [] ) {
+	public function rest_api( string $namespace, callable|string $callback, callable|array|string $args = [] ) {
```
renatonascalves, line 575: Missing `@return`phpDoc and return type.

src/mantle/http/routing/class-url-generator.php
```diff
@@ -211,23 +256,43 @@ public function get_root_url(): string {
...
 	}
 
+	/**
+	 * Get the base URL for the request.
+	 *
+	 * @param  string      $scheme
+	 * @param  string|null $root
+	 * @return string
+	 */
+	public function format_root( string $scheme, ?string $root = null ) {
```
renatonascalves, line 275:
```suggestion
	public function format_root( string $scheme, ?string $root = null ): string {
```

src/mantle/http/routing/class-url-generator.php
```diff
@@ -250,7 +315,7 @@ public function is_valid_url( $path ): bool {
 	 * @param  string $scheme
 	 * @return void
 	 */
-	public function force_scheme( $scheme ) {
+	public function force_scheme( string $scheme ) {
```
renatonascalves, line 318:
```suggestion
	public function force_scheme( string $scheme ): void {
```

src/mantle/http/routing/class-url-generator.php
```diff
@@ -250,7 +315,7 @@ public function is_valid_url( $path ): bool {
 	 * @param  string $scheme
 	 * @return void
```
renatonascalves, line 316: See my previous suggestion about `void`.

src/mantle/support/traits/trait-enumerates-values.php
```diff
@@ -188,23 +188,21 @@ function ( $item ) use ( $key, $value ) {
 	 * @return void
 	 */
 	public function dd( ...$args ) {
```
renatonascalves, line 190:
```suggestion
	public function dd( ...$args ): void {
```

src/mantle/support/traits/trait-enumerates-values.php
```diff
@@ -188,23 +188,21 @@ function ( $item ) use ( $key, $value ) {
...
+		exit( 1 );
 	}
 
 	/**
 	 * Dump the items.
 	 *
-	 * @return $this
+	 * @return static
 	 */
 	public function dump() {
```
renatonascalves, line 201:
```suggestion
	public function dump(): static {
```

src/mantle/testing/class-test-case.php
```diff
@@ -39,6 +39,8 @@
 
 /**
  * Root Test Case for Mantle sites.
+ *
+ * Not designed for external use. Use {@see Mantle\Testkit\Test_Case} instead.
```
renatonascalves, line 43: I can see a dev using this by mistake, especially if they have some IntelliSense in the editor where while importing the class, this note will not appear.

Would you consider adding this as part of custom namespace? Something like `Mantle\Testkit\Internal\Test_Case` or `Internal_Test_Case`...

renatonascalves (approved):
🗡️

src/mantle/testing/class-test-case.php
srtfisher: I can, I will rename it in a follow up

PR #397, opened 2023-06-09
Allow dynamic instance of a model to be created without defining the model class

srtfisher:
As a bridge to defining your own model for a post type, you can use Mantle models to interface with a dynamic post type name. 

The main use case for this PR is for simplified use outside of the framework -- where you want to use Mantle's models and query builder without setting up your own model classes.

```php
Post::for( 'my-post-type' )->where( 'title', 'Title here' )->first();
```


attackant (approved):
🏎️

PR #402, opened 2023-06-23
Improve the handling of various arguments to with_terms()

srtfisher:
Improves the argument handling for those arguments passed to `with_terms()`.

Updates `with_terms()` to support any combination of an array of terms/IDs, term objects, taxonomy => slugs, etc.

Resolves https://github.com/alleyinteractive/mantle/issues/162

### DIFF
--- a/src/mantle/support/class-collection.php
+++ b/src/mantle/support/class-collection.php
@@ -751,7 +751,7 @@ class Collection implements ArrayAccess, Enumerable {
 	 * @template TMapToDictionaryKey of array-key
 	 * @template TMapToDictionaryValue
 	 *
-	 * @param  callable(TValue, TKey): array<TMapToDictionaryKey, TMapToDictionaryValue>|null $callback
+	 * @param  callable(TValue, TKey): array<TMapToDictionaryKey, TMapToDictionaryValue> $callback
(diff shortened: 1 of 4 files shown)


src/mantle/support/class-collection.php
```diff
@@ -1005,12 +1005,21 @@ public function random( $number = null ) {
...
+	 * @template TReduceInitial
+	 * @template TReduceReturnType
+	 *
+	 * @param  callable(TReduceInitial|TReduceReturnType, TValue, TKey): TReduceReturnType $callback
+	 * @param  TReduceInitial                                                              $initial
+	 * @return TReduceReturnType|TReduceInitial
 	 */
 	public function reduce( callable $callback, $initial = null ) {
-		return array_reduce( $this->items, $callback, $initial );
+		$result = $initial;
```
attackant, line 1016: Looks like you could remove this:
```
public function reduce( callable $callback, $initial = null ) {
	
		foreach ( $this as $key => $value ) {
			$initial = $callback( $initial, $value, $key );
		}

		return $initial;
	}
```

attackant (approved):
Made a comment, but otherwise looks good 👍🏼

PR #403, opened 2023-06-23
Fix generator namespaces

srtfisher:
Fix the namespaces in the stubs (`App\Models` vs `App\Models\Models`).


renatonascalves (approved):
🗡️

PR #405, opened 2023-06-30
Feature/dependencies test trait

anubisthejackle:
This trait adds assertions to allow for easily validating whether dependency plugins are loaded during testing.

For the purposes of this PR, I'm defining a "dependency plugin" as a plugin that is active on the Production site that we want to verify is loaded during integration tests to account for interactivity issues between the plugin and the site custom code.

### DIFF
--- a/src/mantle/testing/concerns/trait-dependency-assertions.php
+++ b/src/mantle/testing/concerns/trait-dependency-assertions.php
@@ -1,0 +1,55 @@
+<?php //phpcs:disable WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
+/**
+ * Dependency_Assertions trait file
+ *
+ * @package Mantle
+ */
+
+namespace Mantle\Testing\Concerns;
+
+use PHPUnit\Framework\Assert as PHPUnit;
+use Mantle\Support\Collection;
+
+trait Dependency_Assertions {
+
+	public static function assertDependenciesLoaded( array $dependencies ) {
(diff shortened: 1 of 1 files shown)


GDKeller (approved):
🦖

src/mantle/testing/concerns/trait-dependency-assertions.php
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Asserts that the contents of an array are loaded prior to testing.
+	 *
+	 * Array can be an array of strings or an array of arrays where the first value is a string -- for instance,
+	 * a data provider.
+	 *
+	 * @param array $dependencies Dependencies array.
+	 */
+	public static function assertDependenciesLoaded( array $dependencies ) {
```
renatonascalves, line 23:
```suggestion
	public static function assertDependenciesLoaded( array $dependencies ): void {
```

src/mantle/testing/concerns/trait-dependency-assertions.php
```diff
@@ -0,0 +1,68 @@
...
+			self::assertDependencyLoaded( $dependency );
+		}
+	}
+
+	/**
+	 * Asserts that a file is loaded.
+	 *
+	 * @param string $dependency The dependency file name.
+	 */
+	public static function assertDependencyLoaded( string $dependency ) {
```
renatonascalves, line 47:
```suggestion
	public static function assertDependencyLoaded( string $dependency ): void {
```

PR #409, opened 2023-07-23
Fix http-client content_type method

nlemoine:
Using `content_type()` multiples times (as with `as_form()`, since the object is created with default `as_json()`) results in a header array value which trigger an "array to string conversion" error in Requests


srtfisher (approved):
Thanks for the fix!

PR #413, opened 2023-07-26
Ensure that REST API headers are persisted when testing

srtfisher:
Ensure that REST API headers that were previously ignored are now fully testable.


GDKeller (approved):
💺

PR #416, opened 2023-07-29
Adding chunk()/chunk_by_id()/each()/each_by_id() methods to the query builder, fixing order by aliases

srtfisher:
- Introduces query clause modification for an easy way to modify a query with query builder. In a future state, it'd be nice to do this fluently with `where()` but we'll get there. Supports posts and term queries.
- Fixes some assorted bugs found with order by.
- Adds updated doc blocks to models.


attackant (approved):
Looks good to me 👍🏼

PR #417, opened 2023-08-02
Refactor Factories for simplicity

srtfisher:
Refactors the factory system for simplicity and to mirror Laravel's format for it. 

Also, combines factories for better maintenance. Previously we supported a factory shim for core's test suites and another for Mantle seeding. Now those factories are one an the same. For example:

```php
// In a test...
static::factory()->post->create_and_get();

// Using a model.
App\Model\Post::factory()->create_and_get(); // Returns an instance of App\Model\Post
```

Supports models having custom factories OR rolling with the ones built into Mantle.


attackant (approved):
Looks good Sean, I did not see any obvious errors or issues in the code. If you wanted a higher level dev who knows Mantle better to look at it too, I won't be offended but from what I can tell it looks 👍🏼

PR #421, opened 2023-08-04
User: align user factory `create` method

renatonascalves:
Fix an issue where Mantle is trying to be helpful but changing the behavior of the test when compared using the `WP_UnitTestCase` test class.


scottnelle (approved):
👻

PR #423, opened 2023-08-04
Add dump/dumpSql/dd/ddSql to the query builder

srtfisher:


attackant (approved):
Gonna be dumping SQL all over the place 👍🏼

PR #426, opened 2023-08-14
Add testing for prefer-lowest

### LINKED ISSUE #420: Log_Manager::emergency incompatible with LoggerInterface::emergency
### Description of the bug

When an emergency log message is triggered, this error occurs:

```
PHP Fatal error:  Declaration of Mantle\Log\Log_Manager::emergency(Stringable|string $message, array $context = []): void 
must be compatible with Psr\Log\LoggerInterface::emergency($message, array $context = []) 
in /var/www/wpr/wp-content/vendor/alleyinteractive/mantle-framework/src/mantle/log/class-log-manager.php on line 272
```

This may be due to how we are defining `$message` (the PSR version does not include any types for `$message`) or the fact that we're specifying a return type of `void` when the PSR version doesn't specify a return type at all.

### Steps To Reproduce

Call `Log_Manager::emergency` on PHP >= 8.0 and see the fatal.

### Additional Information

_No response_

srtfisher:
- Add testing for `--prefer-lowest` with Composer to catch any edge cases with outdated dependencies.
- Fixes issues with `psr/log` contract.
- [x] Propagate changes to sub packages.

Fixes #420


mogmarsh (approved):
👍

PR #427, opened 2023-08-15
Support PHPUnit 10

srtfisher:
- [x] Ensure that PHPUnit 10 runs properly.
- [x] Determine path forward for file naming (PHPUnit 10 breaks this).
- [ ] Upgrade `create-wordpress-plugin` to use PSR4 standards for tests.
- [x] Migrate tests in Mantle to PSR4.

For Alleyinz: https://l.alley.dev/b65c9e7aa2


phpunit.xml
```diff
@@ -1,15 +1,15 @@
+<?xml version="1.0"?>
```
renatonascalves, line 1: Does this config works for PHPUnit <10? The way I understood it, it doesn't.

srtfisher: It may or may not but the framework is only using a single version in this PR (~10).

attackant (approved):
🐙

PR #435, opened 2023-08-18
Add better support for a query modifier on a relationship, fix an issue when saving multiple models

srtfisher:


src/mantle/database/model/relations/class-has-one-or-many.php
```diff
@@ -155,34 +156,35 @@ public function add_eager_constraints( Collection $models ): void {
...
+		if ( is_array( $model ) ) {
+			// Return the first model if saving many.
+			return collect( $this->save_many( $model ) )->first();
+		}
+
 		// Save the model if it doesn't exist.
-		if ( ! is_array( $model ) && ! $model->exists && $model instanceof Updatable ) {
+		if ( ! $model->exists && $model instanceof Updatable ) {
+			$model->save();
+		} elseif ( ! $model->exists && $model instanceof Updatable ) {
```
mogmarsh, line 168: this `if` and `elseif` check the same condition
`! $model->exists && $model instanceof Updatable`

mogmarsh (approved):
I flagged one issue that definitely needs attention, otherwise 👍

PR #436, opened 2023-08-18
Adding whereRaw() for modifying the raw query

srtfisher:
Adds `whereRaw()`/`where_raw()` to allow the raw query to be modified. Useful when you need to make a custom query modification and `WP_Query`/`WP_Term_Query` doesn't cut it.


mogmarsh (approved):
👍

PR #440, opened 2023-08-23
Ensure that the console kernel is always booted, not just at the wp-cli command

### LINKED ISSUE #439: Listeners don't bind when using WP-CLI
### Description of the bug

It appears that some listeners, in particular the `Mantle\Database\Model_Service_Provider`'s `on_init` function, do not execute when booting via the CLI kernel. As a result, the model relationship taxonomy is never registered, and anything that depends on it isn't available in a CLI context.

### Steps To Reproduce

1. Create a model relationship using, for example, `belongs_to`.
2. Create a CLI command that programmatically creates a relationship between two posts, e.g., using `->associate()`.
3. Note the error: `Uncaught Mantle\Database\Model\Model_Exception: Error creating internal term for relationship: [Invalid taxonomy.]`

### Additional Information

_No response_

srtfisher:
Ensure that the console kernel is always booted.

Fixes #439.

### DIFF
--- a/src/mantle/database/class-model-service-provider.php
+++ b/src/mantle/database/class-model-service-provider.php
@@ -73,7 +73,7 @@ class Model_Service_Provider extends Service_Provider {
 	/**
 	 * Register the internal taxonomy for post <--> post relationships.
 	 */
-	#[Action('init', 5)]
+	#[Action( 'init', 5 )]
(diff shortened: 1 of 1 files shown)


mogmarsh (approved):
🍿 (kernel, get it?)

PR #447, opened 2023-08-30
Ensure that attribute and action methods are deduplicated

srtfisher:


src/mantle/framework/events/class-discover-events.php
```diff
@@ -80,23 +78,22 @@ protected static function get_listener_events( $listeners, string $base_path ):
...
-								[ $instance->action ],
-								$instance->priority,
-							];
-						}
-
-						continue;
+				// Check if the method has an attribute action.
+				$action_attributes = $method->getAttributes( Action::class );
+
+				if ( ! empty( $action_attributes ) ) {
```
scottnelle, line 84: I assume we know that the `getAttributes()` method will always return an iterable so we don't have to check the type before running `foreach` on it. But if that's not safe to assume, recommend checking the type here.

src/mantle/support/class-service-provider.php
```diff
@@ -108,30 +120,42 @@ function( string $method ) {
...
 
 		foreach ( $class->getMethods() as $method ) {
-			$action_attributes = $method->getAttributes( Action::class );
-
-			if ( empty( $action_attributes ) ) {
-				continue;
-			}
-
-			foreach ( $action_attributes as $attribute ) {
+			foreach ( $method->getAttributes( Action::class ) as $attribute ) {
```
scottnelle, line 145: same here

scottnelle (approved):
👻

PR #450, opened 2023-09-06
Fixing issue with custom post type/taxonomies and factories

srtfisher:
Ensure that when a custom post type/taxonomy uses a factory that the factory uses the proper post type/taxonomy for storage.


efuller (approved):
⚡ a lot of 🪄 here, so feel free to get other 👀 on it.

PR #456, opened 2023-09-22
[1.x] Queue Overhaul

srtfisher:
Overhauling the queue to make it more performant.

- Adds locking, job failures, delaying, and job retrying.
- Refactor to dogfood the Mantle models from within the queue provider.
- Supports concurrent jobs with configurable limits. For example, if 100 jobs are dispatched and 10 concurrent batches can be run, 10 cron events will be dispatched to handle the queued jobs.
- Built with PHP 8.1 in mind (enums) and is slated for a 1.0 release feature. The queue package will require 8.1 but the rest of the framework still supports 8.0.

## Remaining:
- [x] Queue UI #458
- [x] Preserve queue items after completion. #472 
- [x] Automatically deleting expired/old queue jobs. #472
- [x] Update to the Mantle config https://github.com/alleyinteractive/mantle/pull/198


srtfisher:
Reviewed in subsequent pull requests.

PR #460, opened 2023-09-28
Adds a trait to easily silence remote requests during testing

srtfisher:
- Adds a trait that can easily silence and prevent remote HTTP requests during testing. Prevents the need to manually call `fake_request()` when porting over a legacy code base.
- Adds some types to `fake_request()`'s arguments to prevent bad arguments from getting through.


mogmarsh (approved):
👍

PR #462, opened 2023-09-29
Ensure factories can be used with data providers

srtfisher:
- Ensure that database factories can be used with data providers. Previously, they couldn't because the Faker instance in the container wasn't instantiated in time to add the Gutenberg block provider.
- Overhaul the factory generics to allow PHPStan to properly detect the model/object/etc. returned when using the factories.
- Adds `MANTLE_IS_TESTING` constant to make detecting testing easier.

Props to @efuller for pointing this out.


attackant (approved):
👍🏼

PR #481, opened 2023-11-30
[1.x] Allow queue jobs to be run in-admin, allow delays for closure jobs

srtfisher:
Screenshot 2023-11-30 at 4 12 12 PM

- Allow queue jobs to be run from the admin page.
- Fix issues with `add_query_arg()` causing the filter to fall off.
- Allow closure jobs to be delayed.


attackant (approved):
🐙

PR #498, opened 2024-02-21
Switch to environmental variables for installing sqlite/mu-plugins/object cache

srtfisher:


src/mantle/testing/class-utils.php
```diff
@@ -252,6 +252,19 @@ public static function env( string $variable, $default ) {
 		return false === $value ? $default : $value;
 	}
 
+	/**
+	 * Retrieve an environment variable and check if it is truthy.
+	 *
+	 * @param string $variable Variable to get.
+	 * @param bool   $default Default value.
```
attackant, line 259:
```suggestion
	 * @param bool   $default Default value used as a fallback.
```

attackant (approved):
🐙

PR #502, opened 2024-03-01
Add bootable trait to support package

srtfisher:
Add the "magic" bootable trait to the support package.

```php
<?php

use Mantle\Support\Attributes\Action;
use Mantle\Support\Traits\Hookable;

class Example_Feature {
  use Hookable;

  public function __construct() {
    $this->register_hooks();
  }

  public function action__init() {
    // This is done on the init hook
  }

  public function action__wp_enqueue_scripts() {
    // This is done on the wp_enqueue_scripts hook
  }

  #[Action('my_custom_hook')]
  public function register_scripts() {
    // This is done on the my_custom_hook hook
  }
}
```


attackant (approved):
🐙

PR #504, opened 2024-03-01
Add core_object() method to factories to have better compatibility with core

srtfisher:
As titled. Found this during migration of a test suite using core test.


attackant (approved):
🐙

PR #505, opened 2024-03-01
Abstract HTTP requests in testing to standalone class

srtfisher:
Abstract the HTTP requests in testing (`$this->get()`) to a standalone class to prevent conflicts when chaining `$this->with_cookies()` and so on. Mirrors a similar move in the HTTP Client for added code clarity.


attackant (approved):
🐙

PR #507, opened 2024-03-05
Allow WordPress authentication to be defined by an attribute

srtfisher:
Allows for WP Authentication, Deprecated Usage, and Incorrect Usage to be defined on a class/method with an attribute during testing.


attackant (approved):
🐙

PR #516, opened 2024-03-14
Switch CSS selector

srtfisher:
Switch package used to power CSS Selector -> XPath to use Symfony's css-selector package which is more inline with our other dependencies.


mboynes (approved):
🚀

PR #521, opened 2024-04-01
Allow the modified date to be set on models and the database factory

srtfisher:
Also added a `dates` attribute for models.


anubisthejackle (approved):
🎸

PR #522, opened 2024-04-02
Test: use object cache services

renatonascalves:


srtfisher:
This one is failing from another reason from core. Going to patch that today.

renatonascalves:
Closing pr since it was only created for testing the new service.

PR #528, opened 2024-04-11
Element assertions and HTML String

srtfisher:


tests/Testing/Concerns/ElementAssertionsTest.php
```diff
@@ -13,50 +16,85 @@ class ElementAssertionsTest extends Framework_Test_Case {
...
-
-		$response->assertElementExistsById( 'test-id' );
-		$response->assertElementExistsById( '#test-id' );
-		$response->assertElementMissingById( 'missing-id' );
-		$response->assertElementMissingById( '.missing-id' );
+		$this->response()
+			->assertElementExistsById( 'test-id' )
+			->assertElementExistsById( '#test-id' )
+			->assertElementMissingById( 'missing-id' )
+			->assertElementMissingById( '.missing-id' );
```
kevinfodness, line 42: I think this should really be `#missing-id`, no?

kevinfodness (approved):
🍣

tests/Testing/Concerns/ElementAssertionsTest.php
srtfisher: `missing-id` really!

PR #531, opened 2024-04-17
Define WP_RUN_CORE_TESTS during tests

srtfisher:
Relates to https://github.com/Automattic/vip-go-mu-plugins/pull/5476


attackant (approved):
🐙

PR #536, opened 2024-04-19
Allow anonymous models to define events

srtfisher:
For use during testing, make it possible for anonymous models to define events.


attackant (approved):
🐙

PR #541, opened 2024-05-03
[1.1] Overhauling the bootloader to be more flexible

srtfisher:
Allow the bootloader to configure an application without any additional arguments. Also, allow the application to be configured through the bootloader. This will allow any application to truly use the framework's features with just:

```php
bootloader()->boot();
```

Previously, this was a goal and worked but not exactly. The queue didn't work well without a kernel and some more configuration. All of this is now behind us and we can call the bootloader in another application and use the queue/database/etc.

With the changes in this PR we can simplify `bootstrap/app.php` to this:

```php
return Bootloader::create()
	->with_kernels(
		console: App\Console\Kernel::class,
		http: App\Http\Kernel::class,
	)
	->with_exception_handler( App\Exceptions\Handler::class )
	->with_routing(
		web: __DIR__ . '/../routes/web.php',
		rest_api: __DIR__ . '/../routes/rest-api.php',
		pass_through: true,
	);
```


src/mantle/framework/bootstrap/class-load-configuration.php
```diff
@@ -137,33 +151,94 @@ protected function get_configuration_directories( Application $app ): array {
...
-			if ( ! $merge || ! $repository->has( $key ) ) {
-				$repository->set( $key, $config );
-				continue;
+		foreach ( $finder as $file ) {
+			$name = basename( $file->getRealPath(), '.php' );
+
+			$dirname     = basename( dirname( $file->getRealPath() ) );
+			$environment = match ( $dirname ) {
+				// Ignore the 'config' directory as an environment.
+				'config', 'configuration' => null,
```
attackant, line 177: Not really sure, but do any of the changes for loading and merging configuration in this file have performance considerations we should think about?

src/mantle/framework/class-bootloader.php
```diff
@@ -68,13 +76,157 @@ public static function set_instance( ?Bootloader $instance = null ): void {
...
+	/**
+	 * Bind the application with the default kernels.
+	 *
+	 * @throws \InvalidArgumentException If the console or HTTP kernel does not implement the correct interface.
+	 *
+	 * @param class-string<Contracts\Console\Kernel>|null $console Console kernel class.
+	 * @param class-string<Contracts\Http\Kernel>|null    $http    HTTP kernel class.
+	 */
+	public function with_kernels( string $console = null, string $http = null ): static {
+		if ( $console && ! in_array( Contracts\Console\Kernel::class, class_implements( $console ), true ) ) {
```
attackant, line 135: Would a `! class_exists( $handler )` also make sense here? Or would this already be caught? Same question for `with_exception_handler` method below.

attackant (changes requested):
Some things for your review.

src/mantle/database/class-model-service-provider.php
```diff
@@ -28,8 +28,12 @@ class Model_Service_Provider extends Service_Provider {
 	 * Register the service provider.
 	 */
 	public function register(): void {
-		Model::set_event_dispatcher( $this->app['events'] );
+	}
```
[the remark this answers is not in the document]
srtfisher: Yep!

PR #544, opened 2024-05-07
[1.1] Overhaul error handling to remove whoops

srtfisher:
Remove whoops (which has been more unwelcome than welcome in a WordPress environment) and instead render exceptions using Symfony's error handler when called for.


attackant (approved):
You have a couple PHPStan warnings to check, but otherwise looks 👍🏼

PR #553, opened 2024-05-14
[1.1] Load the vip-config.php file during initialization

srtfisher:


src/mantle/testing/wordpress-bootstrap.php
```diff
@@ -86,6 +86,11 @@
 	Utils::info( 'No wp-tests-config.php file found, using default configuration.' );
 }
 
+// Attempt to load the vip-config.php file if it exists to play nicely with VIP Go.
```
renatonascalves, line 89: Should it default to the version from VIP?

Similar to how they do it here: https://github.com/Automattic/vip-container-images/blob/939b55a0cc125a814e235c538f2a324b21416f5a/dev-tools/wp-config-defaults.php#L140-L144

renatonascalves: *if it is a VIP project, I mean...

srtfisher: TIL, can do.

PR #557, opened 2024-06-07
[1.1] Adding support for PHPUnit 11

srtfisher:
No changes needed on the framework side. Added the attributes to silence the deprecation warnings but kept the PHPUnit docblocks in place to support testing in PHPUnit 9.


renatonascalves (approved):
🗡️

PR #560, opened 2024-06-13
Adding support for jobs to be dispatch after the response has been sent

srtfisher:
Send the response to the browser and dispatch the response right away after it is sent.


tests/Queue/providers/WordPressCronQueueTest.php
```diff
@@ -327,6 +327,18 @@ public function test_cron_event_count_assertions() {
 
 		$this->assertCronCount( 'example_hook', 5 );
 	}
+
+	public function test_dispatch_after_response() {
+		$_SERVER['__example_job'] = false;
```
attackant, line 332:
```suggestion
		$_SERVER['__odd_job'] = false; // Throws shoe.
```

attackant: jk

attackant (approved):
🐙

PR #568, opened 2024-07-03
Add support for faking specific requests by HTTP method

srtfisher:
- Added support for faking specific HTTP requests by method.
- Added helper for fluently building HTTP sequence responses.


kevinfodness (approved):
🍣

PR #569, opened 2024-07-03
Add support for generating a post with a thumbnail (a real thumbnail)

srtfisher:


src/mantle/database/factory/class-post-factory.php
```diff
@@ -75,6 +79,36 @@ public function with_thumbnail(): static {
...
+	 * @param int    $width  The width of the image.
+	 * @param int    $height The height of the image.
+	 * @param bool   $recycle Whether to recycle the image file.
+	 */
+	public function with_real_thumbnail( string $file = null, int $width = 640, int $height = 480, bool $recycle = true ): static {
+		return $this->with_middleware(
+			function ( array $args, Closure $next ) use ( $file, $width, $height, $recycle ) {
+				$post = $next( $args );
+
+				update_post_meta(
```
renatonascalves, line 95: it is possible one might want to create a real thumbnail but not set it as the post featured image.

renatonascalves: Or is this already possible today?

renatonascalves (changes requested):

src/mantle/database/factory/class-post-factory.php
srtfisher: Yeah they can just use the attachment factory themselves. I documented that

renatonascalves (approved):

PR #607, opened 2024-12-12
Close output buffer when exception occurs during setup

dlh01:
We have a test that makes a `GET` request and expects an exception to be thrown during the `parse_request` action, which fires as part of`Pending_Testable_Request::setup_wordpress_query()`. 

The exception is thrown, but the output buffer started just before then remains open, leading to a PHPUnit "risky test" warning: `Test code or tested code did not (only) close its own output buffers`. This PR makes sure the buffer is closed.


jakewrfoster (approved):
😎 👍

PR #616, opened 2025-01-14
Add a better dump() helper for responses

srtfisher:


tests/Testing/Concerns/MakesHttpRequestsTest.php
```diff
@@ -38,7 +38,7 @@ public function test_get_home() {
 
 	public function test_get_singular() {
 		$post_id = static::factory()->post->create();
-		$this->get( get_permalink( $post_id ) )
+		$this->get( get_permalink( $post_id ) . '?test=asda' )
```
attackant, line 41: looks like maybe something left from development here?

attackant (approved):
Not sure if there is possibly an artifact from your development here in one file. Otherwise, looks good 👍🏼

PR #617, opened 2025-01-14
Allow object-cache.php to not be being loaded locally

### LINKED ISSUE #609: Disable the object cache dropin for local testing
### Description

Locally, Mantle can load the `object-cache.php` dropin file which can impact the running of the unit tests.

### Use Case

I'd propose a helper method, `->without_object_cache()`, that can be chained. It's similar to how I was able to stop it in a [client project](https://l.alley.dev/6130e3d9a9).

```php

->on( 'enable_loading_object_cache_dropin', function ( $retval ): bool {
    if ( defined( 'VIP_GO_APP_ENVIRONMENT' ) && 'local' === \VIP_GO_APP_ENVIRONMENT ) {
        return false;
    }

    return $retval;
} )
```

This example assumes `VIP_GO_APP_ENVIRONMENT` is used, but we could create a custom env or the helper method could accept a callback/validator.

srtfisher:
Adds a `without_local_object_cache()` method to the installation manager that will prevent the `object-cache.php` drop-in from being loaded locally.

Resolves #609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Add a `without_local_object_cache()` method to prevent the `object-cache.php` drop-in from being loaded locally.
+
 ### Changed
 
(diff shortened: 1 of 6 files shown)


renatonascalves (approved):
🌵

PR #630, opened 2025-01-28
Fix the path to load the vip-config file

renatonascalves:
See #628


src/mantle/testing/wordpress-bootstrap.php
```diff
@@ -93,7 +93,7 @@
 	if ( defined( 'WP_CONTENT_DIR' ) && file_exists( WP_CONTENT_DIR . '/vip-config/vip-config.php' ) ) {
 		require_once( WP_CONTENT_DIR . '/vip-config/vip-config.php' );
 	} elseif ( defined( 'ABSPATH' ) && file_exists( ABSPATH . '/wp-content/vip-config/vip-config.php' ) ) {
-		require_once( ABSPATH . '/wp-content//vip-config.php' );
```
renatonascalves, line 96: I assume this was a typo. Since it fails in a client project. And the vip-config.php inside `wp-content/` is not common (only inside `wp-content/vip-config`.

CleanShot 2025-01-28 at 09 49 37@2x

srtfisher (approved):

PR #637, opened 2025-03-05
Fix issue with model::for() not actually creating a post/term for the post type/taxonomy

srtfisher:
Fixes an issue where `Post::for($post_type)` wouldn't create a post for an underlying post in that custom post type. This was due to a limitation with PHP's anonymous classes. I did spend a lot of time trying to find another way to write this that didn't need `eval()` but had no luck. The Mockery PHP package creates classes on-the-fly using `eval()` in a similar fashion to create mocks.


kingkool68 (approved):
👍

PR #645, opened 2025-03-19
Add mixed helper

srtfisher:
As titled.

### DIFF
--- a/src/mantle/support/class-mixed-data.php
+++ b/src/mantle/support/class-mixed-data.php
@@ -38,7 +38,7 @@ class Mixed_Data implements ArrayAccess, Jsonable, \JsonSerializable, \Stringabl
 	 * Constructor
 	 *
 	 * @param mixed $value Value.
-	 * @param bool  $throw Whether to throw an exception if the value is not a compatible type.
+	 * @param bool  $throw Optional. Whether to throw an exception if the value is not a compatible type.
 	 */
 	public function __construct( mixed $value, bool $throw = false ) {
 		$this->throw = $throw;
--- a/src/mantle/support/class-mixed-data.php
+++ b/src/mantle/support/class-mixed-data.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Mixed_Data class file
+ *
+ * @package mantle-framework
+ */
+
+namespace Mantle\Support;
+
+use ArrayAccess;
+use Mantle\Contracts\Support\Jsonable;
+
+/**
+ * Fluent class for manipulating mixed data as type-safe objects.
+ */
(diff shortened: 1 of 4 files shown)


src/mantle/support/class-mixed-data.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Fluent class for manipulating mixed data as type-safe objects.
+ */
+class Mixed_Data implements ArrayAccess, Jsonable, \JsonSerializable, \Stringable {
+	use Interacts_With_Data;
+
+	/**
+	 * Create a new instance of the class from mixed data.
+	 *
+	 * @param mixed $data Mixed data.
```
renatonascalves, line 22: missing `@return`

src/mantle/support/class-mixed-data.php
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	public static function of( mixed $data ): static {
+		return new static( $data );
+	}
+
+	/**
+	 * Create a new instance of the class.
+	 *
+	 * @param mixed $value Value.
```
renatonascalves, line 31: here too.

src/mantle/support/class-mixed-data.php
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	public static function create( mixed $value ): static {
+		return new static( $value );
+	}
+
+	/**
+	 * Constructor
+	 *
+	 * @param mixed $value Value.
+	 * @param bool  $throw Whether to throw an exception if the value is not a compatible type.
```
renatonascalves:
```suggestion
	 * @param bool  $throw Optional. Whether to throw an exception if the value is not a compatible type.
```

src/mantle/support/helpers/helpers-meta-data.php
```diff
@@ -59,3 +60,12 @@ function user_meta( int $user_id, string $meta_key ): Object_Metadata {
 function comment_meta( int $comment_id, string $meta_key ): Object_Metadata {
 	return Object_Metadata::of( 'comment', $comment_id, $meta_key );
 }
+
+/**
+ * Manage and manipulate mixed data in a type-safe manner.
+ *
+ * @param mixed $value Value to be wrapped.
```
renatonascalves, line 67: Here too.

renatonascalves (changes requested):

src/mantle/support/class-mixed-data.php
srtfisher: This framework does not use @returns with typed returns.

renatonascalves (approved):
🌵

Issue #624, opened 2025-01-15
Upgrade project to PHPStan Level 8


srtfisher:
The project will need types added throughout to get to PHPStan level 8.

With #664 getting us to level 6, 1830 errors remain for level max.

PR #654, opened 2025-04-07
Increasing type level

srtfisher:
Supports #624 to move towards getting this project running with PHPStan level max and type declarations throughout.

### DIFF
--- a/src/mantle/http/trait-interacts-with-input.php
+++ b/src/mantle/http/trait-interacts-with-input.php
@@ -28,17 +28,16 @@ trait Interacts_With_Input {
 	 * @param  string|array|null $default
 	 * @return string|array|null
 	 */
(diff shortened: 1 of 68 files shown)


src/mantle/http/trait-interacts-with-input.php
```diff
@@ -312,12 +306,12 @@ public function cookie( $key = null, $default = null ) {
 	/**
 	 * Retrieve a parameter item from a given source.
 	 *
-	 * @param  string            $source
-	 * @param  string            $key
-	 * @param  string|array|null $default
+	 * @param  string $source
+	 * @param  string $key
```
kingkool68, line 310: Should be `string|null`

kingkool68 (approved):
You missed one 

👍

PR #658, opened 2025-04-07
Remove method name actions when attributes are used

### LINKED ISSUE #657: When using hookable trait, attributes should take priority over method name
When using the `Hookable` trait, attributes should take priority over method names. We do have tests for this right now but it can be unclear to the end user.

```php
/**
 * Enqueues the script.
 */
#[Action( 'wp_head', 5 )]
public function action__wp_head(): void {
```

This method will register the `action__wp_head` action at priority 5 from the attribute and 10 from the method name. The method name should be ignored in this case and only the attribute should be used. This is a departure from what Mantle previously did but the code here isn't clear that it is doubling up on registration.

srtfisher:
Fixes #657


renatonascalves (approved):
⚡

PR #661, opened 2025-04-09
Increasing type level to 40

srtfisher:


benpbolton (approved):
Nice work! Always nice to see a net-negative diff on a change like this. I suppose the adoption of constructor property promotion eliminates a lot of boilerplate code? 

The true tests will be in the testing, but this looks like a great thing!

rector.php
```diff
@@ -43,7 +47,7 @@
 		deadCode: true,
 		instanceOf: true,
 	)
-	->withTypeCoverageLevel( 18 ) // Out of 49.
+	->withTypeCoverageLevel( 30 ) // Out of 49.
```
renatonascalves: how does this relates to phpstan level? is it equivalent?

srtfisher: By increasing the type level in Rector it adds types which makes the upgrade to PHPStan max easier. Once I get to 49 I will start upgrading PHPStan (which is mostly adding generics 😢 )

PR #669, opened 2025-04-17
Adding assertBlockExists/assertBlockMissing to response assertions

srtfisher:


renatonascalves (commented):
So, this is not quite what I had in mind. And oddly, I'm sort of opposed to this approach only because the rendered version of a block, particularly custom blocks, can have any shape and none of the `wp-block-` classes used in the query selector here.

It's also not that different from a `assertSeeText` check.

It's also not quite fluent as using `wp_match` validators. I actually assumed, or hope, it was possible to keep a reference from the `$content` block structures so that it could be used in `$this->get(` requests.

But I can see it'd be hard to parse the rended content back into the block structure to fluently run checks against them.

srtfisher:
@renatonascalves I hear you but I don't know how I would be able to check on a block being in the rendered content in the same way as wp-match-blocks. I think for most decent blocks it should work okay for blocks that use the wrapper attributes. If you have any ideas please feel free to share them 😆

renatonascalves:
@srtfisher Agree! I don't have any ideas _right now_. But I'll try to look at mantle internals to confirm if this idea of mine is _crazy_ or just wishful thinking.

If you don't see a way, probably there isn't one. 🤣

PR #670, opened 2025-04-18
Allow on-the-fly Blade rendering

srtfisher:
Allows for Blade templating to be rendered by passing a Blade template string. Not designed for everyday use with the compiled .blade.php template being the preferred approach.

### DIFF
--- a/src/mantle/http/view/class-view-finder.php
+++ b/src/mantle/http/view/class-view-finder.php
@@ -243,7 +243,7 @@ class View_Finder {
 	 * Check if the view name has hint information.
 	 *
 	 * The hint information is used to determine if the view should be loaded from
-	 * a specific in the format of "@alias/view-name".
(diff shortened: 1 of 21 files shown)


src/mantle/http/view/class-view-finder.php
```diff
@@ -237,4 +238,16 @@ public function get_possible_view_files( string $name ): array {
 			$this->extensions
 		);
 	}
+
+	/**
+	 * Check if the view name has hint information.
+	 *
+	 * The hint information is used to determine if the view should be loaded from
+	 * a specific in the format of "@alias/view-name".
```
kingkool68: A specific what?

kingkool68 (approved):
👍

src/mantle/http/view/class-view-finder.php
srtfisher: 😲

PR #679, opened 2025-04-23
Add fluent post type/taxonomy argument generators

srtfisher:


src/mantle/support/registration/class-post-type-arguments.php
```diff
@@ -0,0 +1,522 @@
...
+	 *
+	 * @param string       $name The name of the method called.
+	 * @param array<mixed> $arguments The arguments passed to the method.
+	 */
+	public function __call( string $name, array $arguments ): static {
+		if ( empty( $arguments ) ) {
+			throw new \InvalidArgumentException( 'No arguments provided for ' . $name );
+		}
+
+		$this->arguments[ $name ] = $arguments[0];
```
mogmarsh, line 518: why `$arguments[0]` and not `$arguments`?

src/mantle/support/registration/class-post-type-arguments.php
```diff
@@ -0,0 +1,522 @@
...
+		if ( ! $plural ) {
+			$plural = Str::plural( $singular );
+		}
+
+		$this->arguments['labels'] = array_merge(
+			$this->arguments['labels'],
+			[
+				'name'                     => $plural,
+				'singular_name'            => $singular,
+				/* translators: %s: Post type name. */
```
mogmarsh: on the taxonomy, you specified Singular or Plural in the translators comment. That would probably be helpful here too, but not a blocker.

src/mantle/support/registration/class-taxonomy-arguments.php
```diff
@@ -0,0 +1,422 @@
...
+				/* translators: %s: Singular label for the taxonomy. */
+				'view_item'                  => sprintf( __( 'View %s' ), $singular ),
+				/* translators: %s: Singular label for the taxonomy. */
+				'update_item'                => sprintf( __( 'Update %s' ), $singular ),
+				/* translators: %s: Singular label for the taxonomy. */
+				'add_new_item'               => sprintf( __( 'Add New %s' ), $singular ),
+				/* translators: %s: Singular label for the taxonomy. */
+				'new_item_name'              => sprintf( __( 'New %s Name' ), $singular ),
+				/* translators: %s: Singular label for the taxonomy. */
+				'template_name'              => sprintf( __( '%s Archives' ), $singular ),
```
mogmarsh: is this one right?

src/mantle/support/registration/class-taxonomy-arguments.php
```diff
@@ -0,0 +1,422 @@
...
+	 *
+	 * @param string       $name The name of the method called.
+	 * @param array<mixed> $arguments The arguments passed to the method.
+	 */
+	public function __call( string $name, array $arguments ): static {
+		if ( empty( $arguments ) ) {
+			throw new \InvalidArgumentException( 'No arguments provided for ' . $name );
+		}
+
+		$this->arguments[ $name ] = $arguments[0];
```
mogmarsh, line 418: same question as before.

tests/Database/Model/Registration/RegisterPostTypeTest.php
```diff
@@ -98,6 +111,31 @@ public function test_field_registration() {
...
 
+class Testable_Registration_Post_Type extends Post implements Registrable {
+	use Register_Post_Type;
+
+	public static $object_name = 'test-post-type';
+
+	public static function get_registration_args(): array {
+		return array(
+			'public' => true,
+		);
```
mogmarsh:
```suggestion
		return [
			'public' => true,
		];
```

mogmarsh (approved):
a couple little questions - probably me not understanding. otherwise 👍

src/mantle/support/registration/class-post-type-arguments.php
srtfisher: you'd end up with an array as the property value (`show_in_ui` would be `[true]` rather than `true`)

PR #684, opened 2025-04-24
Fix header/footer in HTTP requests, ensure that styles/scripts are properly output each run

### LINKED ISSUE #682: Calls to wp_add_inline_script() do not output the inline script when fired on multiple tests
### Description of the bug

When `wp_add_inline_script()` is used to add an inline script, the inline script isn't being outputted on the second test case after it displays on the first. This could be from not cleaning up a WordPress global or something else. This is a fix for the test right now:

```php
	protected function setUp(): void {
		parent::setUp();

		$GLOBALS['wp_scripts']->done = [];
	}
```

### Steps To Reproduce

1. Use `wp_add_inline_script()` to add an inline script across multiple test cases (make it in `wp_head` for example).
2. Make two test cases both with `$this->get( '/' ).
3. In the second test case the script does not appear.

### Additional Information

_No response_

### LINKED ISSUE #683: Missing header/footer in response content when 2+ HTTP requests are made in a test
### Description of the bug

When making a HTTP request in a test with `$this->get( '/' )` the second request never has a header/footer. This is because `get_header()` and `get_footer()` use `locate_template()` with the `$load_once` argument set to true.

```php
$this->get( '/' ); // <!DOCTYPE html><html><body><div>the page</div></body></html>

// The second one:
$this->get( '/' ); // <div>the page</div>
```

### Steps To Reproduce

1. Make a HTTP request: `$this->get( '/' );`
2. Make another and see that the header/footer are missing.

### Additional Information

_No response_

srtfisher:
Resolves #683 
Resolves #682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ensure that multiple HTTP request calls properly load the
   header/footer/sidebar template. Previously only the first request would load
   the header/footer/sidebar templates and the response returned was only the
-  template content .
+  template content.
 
 ## v1.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   directory and from `routes/console.php` if it exists without needing to define
   a `commands()` method in your application's console kernel.
 
+### Fixed
+
+- Ensure that multiple HTTP request calls properly load the
+  header/footer/sidebar template. Previously only the first request would load
+  the header/footer/sidebar templates and the response returned was only the
+  template content .
+
 ## v1.6.0
 
 ### Added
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   directory and from `routes/console.php` if it exists without needing to define
   a `commands()` method in your application's console kernel.
 
+### Fixed
+
+- Ensure that multiple HTTP request calls properly load the
+  header/footer/sidebar template. Previously only the first request would load
+  the header/footer/sidebar templates and the response returned was only the
+  template content .
```
renatonascalves:
```suggestion
  template content.
```

tests/Testing/Concerns/MakesHttpRequestsWithTemplatesTest.php
```diff
@@ -0,0 +1,56 @@
...
+use function Mantle\Testing\iterate_test;
+
+/**
+ * Tests for making HTTP requests in unit tests that relate to cleaning up
+ * globals (such as enqueued scripts) and ensuring that requests have a
+ * header/footer.
+ *
+ * @group testing
+ */
+#[Group( 'testing' )]
```
renatonascalves, line 19: duplicate?!

renatonascalves (approved):
🌵

tests/Testing/Concerns/MakesHttpRequestsWithTemplatesTest.php
srtfisher: yes we support phpunit 9-11

PR #705, opened 2025-06-04
Restore model event dispatcher

### LINKED ISSUE #701: Exception when using model events
### Description of the bug

When using model events, I hit an exception:

```
PHP Fatal error:  Uncaught InvalidArgumentException: Model event dispatcher not set: [App\Models\My_Model] in [...]/mantle-framework/src/mantle/database/model/concerns/trait-has-events.php:38
```

It looks like #541 [removed the following line from the Model Service Provider](https://github.com/alleyinteractive/mantle-framework/blame/d47209e3816f6de6f7a91a958b023725068aef9d/src/mantle/database/class-model-service-provider.php#L31C3-L31C55):

```php
Model::set_event_dispatcher( $this->app['events'] );
``` 

### Steps To Reproduce

1. Create a model
2. In the boot method for the model, add a model event listener:
```php
protected static function boot() {
    static::created( function( $post ) {
        // ...
    } );
}
```
3. Experience the error.

### Additional Information

I believe our codebase is using a slightly outdated version of mantle, so this should be confirmed on the latest one.

srtfisher:
Fixes #701

### DIFF
--- a/src/mantle/database/class-model-service-provider.php
+++ b/src/mantle/database/class-model-service-provider.php
@@ -27,13 +27,14 @@ class Model_Service_Provider extends Service_Provider {
 	/**
 	 * Register the service provider.
 	 */
-	public function register(): void {
-	}
+	public function register(): void {}
 
 	/**
 	 * Bootstrap the service provider.
 	 */
 	public function boot(): void {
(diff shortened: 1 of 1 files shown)


kingkool68 (approved):
👍

Issue #660, opened 2025-04-09
[2.x] Refactor queue to use custom database tables


srtfisher:
### Description

I want to be able to use custom database tables for queue jobs instead of polluting wp_posts. This could make it more performant, too.

### Use Case

When a developer uses the queue the job data should live within a custom database table.

kevinfodness:
@srtfisher any update here?

srtfisher:
I'm going to call this done because it'll be in 2.0.

PR #707, opened 2025-06-05
Add a database table model and query builder

srtfisher:
Supports #660 by setting up a way to interface with a custom database table.


kingkool68 (approved):
👍

PR #708, opened 2025-06-06
[2.x] Moving queue to standalone tables for better performance and isolation.

### LINKED ISSUE #660: [2.x] Refactor queue to use custom database tables

srtfisher:
Resolves #660


srtfisher:
Folded in a follow-up commit (`74f081ac`) with correctness fixes from a review pass on top of the feature:

- **Atomic job claiming + crash recovery** — `pop()` now claims each job via a status-guarded `UPDATE … WHERE id = ? AND status IN ('pending','running') AND available_at_gmt <= now`, so concurrent workers can never pop the same job, and jobs left `running` by a crashed worker are reclaimed once their lock expires.
- **`Worker::run_single()` restored to public** — the admin "Run" action calls it externally and would otherwise fatal.
- **`queue:cleanup` name/signature untangled** — `--legacy` is now a real, working flag (previously `$name` embedded the option and `$signature` declared an unused `--skip-legacy`).
- **`push()`** resolves queue/delay from the job *before* serializing a closure (avoids reading a property on a string).
- **`Has_Attributes::from_json()`** restores `JSON_THROW_ON_ERROR` so malformed JSON isn't silently swallowed framework-wide.
- Fixed the delete-button confirm text ("delete" not "retry") and added two tests covering expired-lock reclaim / live-lock protection.

Verified locally: `--group queue` 24 passed / 1 intentional incomplete; `WorkerTest` + `InteractsWithCronTest` 4 passed.

PR #715, opened 2025-06-16
Adding PreserveObjectCache attribute

### LINKED ISSUE #711: Allow request testing without clearing the cache
### Description

When testing a request using https://mantle.alley.com/docs/testing/requests#making-requests, the cache is cleared before each request. This makes it difficult to test something in the request that retrieves data from the cache.

https://github.com/alleyinteractive/mantle-framework/blob/9bdf107de5d9da76eba01df35344831dc8b0eb0b/src/mantle/testing/class-pending-testable-request.php#L578

## Example test to show the problem
```php
// Set some cache.
wp_cache_set( 'key', 'example' );

$this->get( '/....' )->assertContains( 'example' ); // The cache is flushed before the request is made. This will fail.

wp_cache_get( 'key' ); // false
```

Consider making this opt-in with a trait or the default (?). This has been the way it has been since the start of the project.

### Use Case

When a test request is made when testing, the cache will not be cleared after each request.

srtfisher:
Resolves #711

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Adds a `PreserveObjectCache` attribute to prevent the object cache from being
+  cleared during unit test HTTP requests.
 - Added `Database_Table_Model` class to allow for creating models that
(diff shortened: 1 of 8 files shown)


kingkool68 (approved):
👍

PR #720, opened 2025-07-07
Adjust how nunomaduro/collision is loaded in unit tests

srtfisher:
Also prevents it from being loaded when phpunit is not being invoked.


kingkool68 (approved):
👍

PR #722, opened 2025-07-10
Suppress errors during blog/site creation to prevent errors when creating temporary tables

srtfisher:
Props to @renatonascalves


renatonascalves (approved):
🥳

PR #732, opened 2025-07-25
Enhance HTTP pool requests to have parity with non-pooled requests

srtfisher:
```php
use Mantle\Facade\Http;

Http::pool( fn ( Pool $pool ) => [
	$pool->get( 'https://example.com/async/' ),
	$pool->post( 'https://example.com/second-async/' ),
] );
```

### DIFF
--- a/src/mantle/http-client/class-pending-request.php
+++ b/src/mantle/http-client/class-pending-request.php
@@ -9,27 +9,30 @@
 
 namespace Mantle\Http_Client;
 
-use DateTimeInterface;
 use InvalidArgumentException;
 use Mantle\Support\Pipeline;
 use Mantle\Support\Traits\Conditionable;
 use Mantle\Support\Traits\Macroable;
 
-use function Mantle\Support\Helpers\collect;
 use function Mantle\Support\Helpers\retry;
 use function Mantle\Support\Helpers\tap;
 
 /**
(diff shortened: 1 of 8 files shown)


src/mantle/http-client/class-pending-request.php
```diff
@@ -584,16 +574,11 @@ public function head( string $url, array|string|null $query = null ): Response {
 	/**
 	 * Issue a POST request to the given URL.
 	 *
-	 * @throws InvalidArgumentException If the request is pooled.
-	 *
 	 * @param  string                    $url URL to post.
 	 * @param  array<string, mixed>|null $data Data to send with the request.
+	 * @return Response
```
renatonascalves, line 579: isn't that against your own rule of not adding `@return` to this project?

renatonascalves: the `mixed` type returned also does not match. 🤔

renatonascalves (approved):
🌵

src/mantle/http-client/class-pending-request.php
srtfisher: It's to make the extended Pooled Pending Request class happy and the IDE aware that it's returning a response. It's not ideal and I am working on refactoring the whole thing to make it better for 2.0.

PR #735, opened 2025-08-01
Allow a factory to be passed as model meta

srtfisher:


mogmarsh (approved):
👍 very nice. I think we could use this.

PR #745, opened 2025-08-19
Improve performance of dump() method, add information about current query

srtfisher:
The performance of termwind's `<code>` block was too slow for large HTML responses.

### DIFF
--- a/src/mantle/support/traits/trait-enumerates-values.php
+++ b/src/mantle/support/traits/trait-enumerates-values.php
@@ -733,7 +733,7 @@ trait Enumerates_Values {
 	 *
 	 * @return array<TKey, TValue>
 	 */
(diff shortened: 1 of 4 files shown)


src/mantle/support/traits/trait-enumerates-values.php
```diff
@@ -733,7 +733,7 @@ public function toArray() { // phpcs:ignore WordPress.NamingConventions.ValidFun
 	 *
 	 * @return array<TKey, TValue>
 	 */
-	public function jsonSerialize(): mixed { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
+	public function json_serialize(): mixed { // phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
```
kingkool68, line 736: Now the method name is valid

kingkool68 (approved):
👍

PR #754, opened 2025-08-28
Add a memo() helper to memoize values based on dependencies

srtfisher:
Similar to Laravel and Spatie's `once()` method but only uses an explicit set of dependencies rather than all variables passed to the parent method.


kingkool68 (approved):
👍

PR #757, opened 2025-09-02
[1.x] Allow snapshot HTTP request mocking

### LINKED ISSUE #758: Allow snapshot HTTP request mocking
From @kevinfodness 

> Suggestion: I love Mantle's ability to mock/intercept HTTP requests to avoid making actual HTTP requests during tests. However, setting that up can be semi-onerous, particularly if what you're trying to do is to cache an oEmbed response or something. It would be really useful to have a way of saying "I want this request to be cached, so make the request and then cache the response in a way that can be committed to source control and used in the future when running this test."

srtfisher:
Implements #758


src/mantle/testing/concerns/snapshots/trait-mocks-http-requests.php
```diff
@@ -0,0 +1,194 @@
...
+	 *
+	 * @param Request $request
+	 */
+	private function get_snapshot_id( Request $request ): string {
+		$test_case = $this->get_test_case();
+
+		return collect( [
+			$test_case->nameWithDataSet(),
+			$request->enum_method()->value,
+			str_replace( [ '/', ':', DIRECTORY_SEPARATOR ], '-', $request->url() ),
```
kevinfodness: This will work for oEmbed requests, but wouldn't work for requests that go to a generic endpoint that accepts arguments (e.g., a REST API or GraphQL endpoint that accepts POST arguments or a JSON payload). Would it be possible to make a hash out of the request (e.g., URL + HTTP verb + headers + body)? This may be trivial to do since under the hood WP is building a request object that has this information in it already - it could be converted to a string and hashed, or the relevant pieces could be JSON-encoded and hashed, or similar. If we want to preserve the ability to identify what tests the snapshot is associated with, we could keep the test case class and method names, and append the hash to the end rather than a representation of the URL.

kevinfodness (approved):
Nice! One suggestion, otherwise 🍣

src/mantle/testing/concerns/snapshots/trait-mocks-http-requests.php
srtfisher: @kevinfodness good call. I have added two things:

1. Header and body MD5s to the default snapshot file name.
2. Custom override (set the snapshot name if you want to do your own thing).

PR #761, opened 2025-09-05
Test traits too

szepeviktor:
Now errors in traits cannot escape PHPStan!

@srtfisher Please give a better path/name than `tests/phpstan/With_Faker_Trait_Test.php`


srtfisher:
@szepeviktor we should follow PSR standards here with the other tests. `WithFakerTraitTest` maybe? what do you think?

szepeviktor:
@srtfisher All done.

PHPStan says these traits depend on classes using them.
Do you know what to do now?

PR #780, opened 2025-09-26
Issue-779: Hookable: add support for validator attributes

### LINKED ISSUE #779: Hookable: add support for validator attributes
### Description

Allows passing a validator attribute to a method and the `Hookable` trait only register hook if it passes validation.

### Use Case

```php
class MyFeature extends Hookable_Feature {
   
   #[MyCustomValidator( true )]
   #[Action('init')]
   public function cool_hook(): void {}

   #[MyCustomValidator( true )]
   public function on_cool_hook(): void {}
}
```

renatonascalves:
### Summary

Fixes #779

Adds support for validator attributes to the `Hookable` trait, enabling conditional registration of hooks based on custom validation logic. Hooks are now only registered if their associated validator attribute passes.

### Description

This PR enhances the `Hookable` trait by introducing support for validator attributes on hook methods. When a method in a class using `Hookable` is decorated with a validator attribute, the hook is only registered if the validator returns `true`. This enables more granular control over which hooks are active, based on custom logic defined in validator attributes.

Key updates include:
- Updates to the `trait-hookable.php` docblocks and logic to document and support validator attributes.
- Addition of a new `TestValidatorAttribute` class for testing, implemented as a PHP attribute that determines hook registration eligibility.
- Expanded test coverage to verify correct behavior when validator attributes are present and return both true and false values.

### Use Case

```php
class MyFeature extends Hookable_Feature {

   #[MyCustomValidator(true)]
   #[Action('init')]
   public function cool_hook(): void {}

   #[MyCustomValidator(true)]
   public function on_cool_hook(): void {}
}
```
With this update, the above hooks will only be registered if `MyCustomValidator` returns `true`.

### Testing Instructions

- Add a validator attribute to a method in a class that uses the `Hookable` trait.
- Confirm that the hook is registered only if the validator condition passes.
- Test with both passing and failing validator conditions to ensure expected behavior.

### Additional Testing Details

- New tests have been introduced in `HookableAttributeTest` and `HookableMethodNameTest` to cover scenarios where validator attributes are active and inactive.
- The `TestValidatorAttribute` is used to simulate both passing and failing validation conditions, confirming that hooks are properly registered or skipped.

### Checklist

- [x] I have tested these changes locally.
- [x] I have added or updated tests as appropriate.
- [ ] I have updated relevant documentation as needed.

### DIFF
--- a/src/mantle/support/traits/trait-hookable.php
+++ b/src/mantle/support/traits/trait-hookable.php
@@ -212,7 +212,7 @@ trait Hookable {
 				}
 
 				// Check if the method passes all validators.
-				if ( false === $this->fire_validator( $method ) ) {
+				if ( false === $this->validate_method( $method ) ) {
 					continue;
 				}
 
--- a/src/mantle/support/traits/trait-hookable.php
+++ b/src/mantle/support/traits/trait-hookable.php
@@ -282,13 +282,6 @@ trait Hookable {
 			if ( ! $attribute->newInstance()->validate() ) {
 				return false;
 			}
-
-		// Check all validators for this method.
-		foreach ( $attributes as $attribute ) {
-			$instance = $attribute->newInstance();
-			if ( $instance instanceof \Mantle\Types\Validator && ! $instance->validate() ) {
-				return false;
-			}
 		}
 
 		return true;
--- a/src/mantle/support/traits/trait-hookable.php
+++ b/src/mantle/support/traits/trait-hookable.php
@@ -133,7 +133,7 @@ trait Hookable {
 					}
 
 					// Check if the method passes all validators.
-					if ( false === $this->fire_validator( $reflection_method ) ) {
+					if ( false === $this->validate_method( $reflection_method ) ) {
 						return null;
 					}
 
@@ -188,7 +188,7 @@ trait Hookable {
 				}
 
 				// Check if the method passes all validators.
-				if ( false === $this->fire_validator( $method ) ) {
+				if ( false === $this->validate_method( $method ) ) {
 					continue;
 				}
 
--- a/src/mantle/support/traits/trait-hookable.php
+++ b/src/mantle/support/traits/trait-hookable.php
@@ -273,7 +273,6 @@ trait Hookable {
 	 * Fire all validators for a method.
 	 *
 	 * @param ReflectionMethod $method The hook callback method.
-	 * @return bool
 	 */
 	protected function fire_validator( ReflectionMethod $method ): bool {
 		$attributes = $method->getAttributes();
@@ -281,10 +280,8 @@ trait Hookable {
 		// Check all validators for this method.
 		foreach ( $attributes as $attribute ) {
 			$instance = $attribute->newInstance();
-			if ( $instance instanceof \Mantle\Types\Validator ) {
-				if ( ! $instance->validate() ) {
-					return false;
-				}
+			if ( $instance instanceof \Mantle\Types\Validator && ! $instance->validate() ) {
+				return false;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


src/mantle/support/traits/trait-hookable.php
```diff
@@ -247,4 +268,26 @@ protected function fire_doing_it_wrong( ReflectionMethod $method ): void {
...
 		_doing_it_wrong( esc_html( static::class . '::' . $method->getName() ), esc_html( $message ), '1.0.0' );
 	}
+
+	/**
+	 * Fire all validators for a method.
+	 *
+	 * @param ReflectionMethod $method The hook callback method.
+	 * @return bool
+	 */
+	protected function fire_validator( ReflectionMethod $method ): bool {
```
renatonascalves: terrible name, open to suggestions.

src/mantle/support/traits/trait-hookable.php
```diff
@@ -247,4 +268,26 @@ protected function fire_doing_it_wrong( ReflectionMethod $method ): void {
...
+		$attributes = $method->getAttributes();
+
+		// Check all validators for this method.
+		foreach ( $attributes as $attribute ) {
+			$instance = $attribute->newInstance();
+			if ( $instance instanceof \Mantle\Types\Validator ) {
```
renatonascalves: We have other validators at alley. Not sure if we want to expand this. But I used the one the framework supports.

src/mantle/support/traits/trait-hookable.php
```diff
@@ -247,4 +268,23 @@ protected function fire_doing_it_wrong( ReflectionMethod $method ): void {
...
 		_doing_it_wrong( esc_html( static::class . '::' . $method->getName() ), esc_html( $message ), '1.0.0' );
 	}
+
+	/**
+	 * Fire all validators for a method.
+	 *
+	 * @param ReflectionMethod $method The hook callback method.
+	 */
+	protected function fire_validator( ReflectionMethod $method ): bool {
+		$attributes = $method->getAttributes();
```
srtfisher: this will limit the attributes for you without having to do an instance of below.

```suggestion
		$attributes = $method->getAttributes( \Mantle\Types\Validator::class, \ReflectionAttribute::IS_INSTANCEOF );

		// Check all validators for this method.
		foreach ( $attributes as $attribute ) {
			if ( ! $attribute->newInstance()->validate() ) {
				return false;
			}
```

src/mantle/support/traits/trait-hookable.php
```diff
@@ -247,4 +268,23 @@ protected function fire_doing_it_wrong( ReflectionMethod $method ): void {
 
 		_doing_it_wrong( esc_html( static::class . '::' . $method->getName() ), esc_html( $message ), '1.0.0' );
 	}
+
+	/**
+	 * Fire all validators for a method.
+	 *
+	 * @param ReflectionMethod $method The hook callback method.
+	 */
+	protected function fire_validator( ReflectionMethod $method ): bool {
```
srtfisher:
```suggestion
	protected function validate_method( ReflectionMethod $method ): bool {
```

src/mantle/support/traits/trait-hookable.php
```diff
@@ -190,11 +205,17 @@ private function collect_attribute_hooks(): Collection {
...
 
 			foreach ( $method->getAttributes( Filter::class ) as $attribute ) {
+				// Skip non-public methods.
 				if ( ! $method->isPublic() ) {
 					$this->fire_doing_it_wrong( $method );
 					continue;
 				}
 
+				// Check if the method passes all validators.
+				if ( false === $this->fire_validator( $method ) ) {
```
srtfisher:
```suggestion
				if ( false === $this->validate_method( $method ) ) {
```

srtfisher (approved):
thanks for the contribution! I haven't documented `mantle-framework/types` yet but I'll include this in it, too.

Issue #699, opened 2025-05-09
HTML: add helper for getting element before/after the current one


renatonascalves:
### Description

Add support to the `\Mantle\Support\HTML` class to get the element **before** or/and **after** the current one.

### Use Case

I needed to get the previous or the after element after filtering them. The `div` was available but there was no quick way around it.

```php
$this->get( $post )
    ->assertOk()
    ->assertQueryTrue( 'is_singular', 'is_single' )
    ->assertBlockExists( 'editor/newsletter', 1 )
    ->html()
    ->filter( 'div.single__content' )
    ->children( 'p,div' )
    ->assertElementExistsByClassName( 'single__inline-module' )
    // Try to get the div before ^ this one.
```

I ended doing something like this:

```php
...
->previousAll()
->filter( 'div' )
->first()
// get the previous element.
```

dustinpaluch:
@renatonascalves next_sibling and previous_sibling have merged into 1.x!

PR #815, opened 2025-10-27
Add previous and next sibling functions

dustinpaluch:
Addresses https://github.com/alleyinteractive/mantle-framework/issues/699


srtfisher:
@dustinpaluch can we get the tests fixed before code reviewing?

dustinpaluch:
@srtfisher sorry about that! Take another look?

src/mantle/support/class-html.php
```diff
@@ -152,6 +152,36 @@ public function first_by_selector( string $selector ): static {
 		return $this->filter( $selector )->first();
 	}
 
+	/**
+	 * Get the nearest previous sibling element.
+	 *
+	 * @param string|null $selector Optional CSS selector to filter the previous siblings.
+	 * @throws \InvalidArgumentException If the current node list is empty.
+	 */
+	public function previousSibling( ?string $selector = null ): static {
```
srtfisher: These should be underscore_cased -- this class does have a mix of them because of the extended class but I was trying to make our methods use underscore_case.

srtfisher (changes requested):

src/mantle/support/class-html.php
dustinpaluch: Fixed!

srtfisher (approved):

PR #832, opened 2025-11-12
Preserve WordPress Globals

### LINKED ISSUE #826: Preserve meta/post types/taxonomies registered before test run
### Description

Mantle is currently taking too aggressive of a stance when cleaning up between test runs. Right now, it [resets all meta keys](https://github.com/alleyinteractive/mantle-framework/blob/72bc20b307db22ec439b61250eeeb46e84682a5a/src/mantle/testing/TestCase.php#L274) in the tear down for each test run. This can cause issues when attempting to use meta that was registered on `init` before the test case even started.

We should respect all existing meta that is registered before the test runs and continue to reset it back after each run.

## WordPress Globals to Consider

- `$wp_meta_keys`: registered meta keys. Currently being reset without any method to prevent it.
- `$wp_post_types`: registered post types. Not reset by default. Only reset if a test case uses the `Reset_Data_Structures` trait.
- `$wp_taxonomies`: registered taxonomies. Not reset by default. Only reset if a test case uses the `Reset_Data_Structures` trait.
- `$wp_widget_factory`: registered widgets. Not reset at all.

For this ticket, we will only consider object meta.

## Acceptance Criteria

1. The `$wp_meta_keys` global is backed up before starting the test.
2. The global is reset during each `setUp` method.
3. An attribute is restore current behavior and prevent the restoration of meta keys between tests (what we have now).

### LINKED ISSUE #827: Preserve and reset post types / taxonomies between test runs
Continuing from https://github.com/alleyinteractive/mantle-framework/issues/826...

> Mantle is currently taking too aggressive of a stance when cleaning up between test runs. Right now, it [resets all meta keys](https://github.com/alleyinteractive/mantle-framework/blob/72bc20b307db22ec439b61250eeeb46e84682a5a/src/mantle/testing/TestCase.php#L274) in the tear down for each test run. This can cause issues when attempting to use meta that was registered on `init` before the test case even started.

We should make it possible to reset post types and taxonomies without unregistering them entirely. This would allow tests to have a clean slate: the same post types/taxonomies (including custom post types/taxonomies) registered between tests.

srtfisher:
Resolves #827, resolves #826.

Taxonomies, post types, and meta keys are now backed up on first load and then restored on setUp. This will allow the framework to provide a better testing environment for end users.


mogmarsh (approved):
👍

PR #856, opened 2026-01-07
Fix issues with clearing out globals that results in orphaned data. Account for sitemaps, feeds, and robots in testing.

### LINKED ISSUE #853: The global $wp_sitemaps gets unset but is still references in core hooks
### Description of the bug

[On `init`](https://github.com/WordPress/WordPress/blob/c900da708a6b998d02e1df79a20633d9193967d9/wp-includes/default-filters.php#L541-L542), WordPress core [initializes the global `$wp_sitemaps`](https://github.com/WordPress/WordPress/blob/c900da708a6b998d02e1df79a20633d9193967d9/wp-includes/sitemaps.php#L26-L28), which then [hooks the class instance into `template_redirect` and `robots_txt` and registers rewrite tags (including query vars) and rules](https://github.com/WordPress/WordPress/blob/c900da708a6b998d02e1df79a20633d9193967d9/wp-includes/sitemaps/class-wp-sitemaps.php#L67-L78).

On `tearDown()`, [Mantle nulls-out the global `$wp_sitemaps`](https://github.com/alleyinteractive/mantle-framework/blob/50fcbbf47c2f3b2889ce0daffd1e9f3f5bb2e0a9/src/mantle/testing/TestCase.php#L274). Because the instance's methods were hooked into `template_redirect` and `robots_txt` on `init`, that class instance gets orphaned, and subsequent calls to `wp_sitemaps_get_server()` will new up an additional `WP_Sitemaps` instance and add further callbacks onto `template_redirect` and `robots_txt`. This can lead to a variety of issues. In my particular use case, I'm modifying the default sitemap renderer (which is stored in a property of the `WP_Sitemaps` instance), and when `template_redirect` fires, the `WP_Sitemaps` that was hooked first runs and exits and the instance that I modified is never touched.

### Steps To Reproduce

In a test, call `wp_sitemaps_get_server()` and observe that the `$GLOBALS['wp_filters']['template_redirect']` now has two `WP_Sitemaps::render_sitemap()` callbacks registered.

### Additional Information

Beyond this specific scenario, this has some further-reaching implications. What other globals are we nulling out, which have callback methods hooked against orphaned class instances?

### LINKED ISSUE #855: Improve testing of feeds and sitemaps
### Description

I would like it to be easier to test core sitemaps AND RSS feeds (built-in and custom ones).

### Use Case

When a user adds a new sitemap or feed, they can test it using the testing framework.

Related #853

srtfisher:
- **Migrating to contract for responses**
- **Adding a test for feeds**

Resolves #855. Resolves #853.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,18 +82,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dependencies: ["prefer-lowest", "prefer-stable"]
+        dependencies: ["prefer-stable"]
         multisite: [true, false]
         # See core's PHP Compatibility chart:
         # https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/
         php: [8.2, 8.3, 8.4, 8.5]
         wordpress: ${{fromJson(needs.get-versions.outputs.versions)}}
+        vip-mu-plugins: [true, false]
         exclude:
-          # Skip testing with prefer-lowest on more recent PHP versions at this time.
-          - php: 8.4
-            dependencies: "prefer-lowest"
-          - php: 8.5
-            dependencies: "prefer-lowest"
           # WordPress 6.7 added support for PHP 8.4.
           - php: 8.4
             wordpress: "6.6"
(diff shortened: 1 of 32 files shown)


mogmarsh (approved):
👍

PR #857, opened 2026-01-13
Ensure rewrite is properly preserved and cleared

### LINKED ISSUE #854: Rewrite rules persist between tests, but tags do not
### Description of the bug

On `tearDown()`, [Mantle replaces the global `WP` instance with a new one](https://github.com/alleyinteractive/mantle-framework/blob/50fcbbf47c2f3b2889ce0daffd1e9f3f5bb2e0a9/src/mantle/testing/TestCase.php#L253). Amongst other things, the global `WP` instance [stores any query vars that get added](https://github.com/WordPress/WordPress/blob/b9a793c0f178bca73ccc02dce18a1738073b2000/wp-includes/class-wp.php#L10-L18), e.g. [via `add_rewrite_tag()`](https://github.com/WordPress/WordPress/blob/b9a793c0f178bca73ccc02dce18a1738073b2000/wp-includes/rewrite.php#L172), which calls [`WP::add_query_var()`](https://github.com/WordPress/WordPress/blob/b9a793c0f178bca73ccc02dce18a1738073b2000/wp-includes/class-wp.php#L86-L97).

The global `WP_Rewrite` instance, it should be noted, is not in [the list of globals](https://github.com/alleyinteractive/mantle-framework/blob/50fcbbf47c2f3b2889ce0daffd1e9f3f5bb2e0a9/src/mantle/testing/TestCase.php#L251-L282) that get newed up or nulled out. Therefore, some code may add rewrite tags and rules, and after `TestCase::tearDown()` runs, the rewrite rule will match a request but not work. I observed this to be the case with core sitemap requests, [which do exactly this](https://github.com/WordPress/WordPress/blob/b9a793c0f178bca73ccc02dce18a1738073b2000/wp-includes/sitemaps/class-wp-sitemaps.php#L131-L153).

### Steps To Reproduce

1. On `init`, add a rewrite tag and a rewrite rule that uses the rewrite tag.
2. In a second test (since `tearDown()` must run at least once before this is experienced), attempt to use the rewrite rule.
3. Observe that the rewrite rule matched the request, but the query var was not usable.

### Additional Information

This relates to #853, as well as #826, #827, #828, #832

srtfisher:
Fixes #854

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **Potentially breaking:** The `$wp_rewrite` global and public query variables in `WP`
+  are now backed up before any tests are run and restored before each test to prevent
+  side effects from tests that modify these globals. This will allow tests to properly
+  modify the rewrite rules and public query variables without affecting other tests.
 - Adds PHPStan testing to more traits.
 - Clear object metadata lazy load queue when tearing down a test.
-- **Potentially breaking:** The `$wp_rewrite` global and public query variables in `WP` are now backed up
-  before any tests are run and restored before each test to prevent side effects
-  from tests that modify these globals. This will allow tests to properly modify
-  the rewrite rules and public query variables without affecting other tests.
 
 ## v1.16.0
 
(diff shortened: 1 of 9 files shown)


kingkool68 (approved):
👍

PR #860, opened 2026-01-27
Adding a Delayed Feature class

srtfisher:


tests/Types/DelayedFeatureTest.php
```diff
@@ -0,0 +1,223 @@
...
+		$delayed_low = new Delayed_Feature(
+			'init',
+			function() {
+				$_SERVER['execution_order'][] = 'low';
+			},
+			15
+		);
+
+		$delayed_high->boot();
+		$delayed_low->boot();
```
mogmarsh: I would switch these two lines, so it's really obvious that the priority is working

mogmarsh (approved):
one suggestion, otherwise 👍

Issue #700, opened 2025-05-09
HTML: improve assertions while chaining


renatonascalves:
### Description

As a new user of the `\Mantle\Support\HTML` class helper methods, I noticed something that can be prone to induce a user to make a mistake.

I wonder if it is important to help users, either with documentation or maybe with errors so that it doesn't happen or so that the user is aware of them.

### Use Case

For example, take this html:

```html
<div class="unrelated-class-not-in-the-random-element"></div>
<div id="random-element"></div>'
```

A user would assume that the `has_class` method is checking the `div` with the id, unaware it is actually checking the whole `string`.
```php
$response->html()
	->filter( 'div' )
        ->eq( 1 )
	->assertElementExistsById( 'random-element' )
	->has_class( 'unrelated-class-not-in-the-random-element' )
```

So the fact that the HTML class has multiple helpers, including methods from `SymfonyCrawler`, makes it prone to mistakes, particularly while writing unit tests where chaining them is so easy.

dustinpaluch:
it seems that it is not checking the whole string, so `eq` is working correctly, and `has_class` is returning false in your example, but as it is not an assertion, and the only other existing assertion passes, it feels like `has_class` is matching when it's not.

I think it's a code smell that this class has methods both for business logic and assertion methods that are only relevant in a unit testing context. I think we should take the assertion traits off of HTML and make them into standalone traits for test classes, but that would be a breaking change. Open to other ideas.

dustinpaluch:
@renatonascalves any thoughts?

renatonascalves:
@srtfisher What do you think?

cobianzo:
Hi, I've opened a PR with a non-breaking fix for this: https://github.com/alleyinteractive/mantle-framework/pull/883 
I did it, well, Claude Code did during the retreat AI session. It's one of the solutions suggested by Dustin (feel free to accept it or dismiss it and put the task back to 'Ready')

  The approach adds four new chainable assertion methods to HTML\Assertions — `assertHasClass()`, `assertNotHasClass()`,
  `assertHasAnyClass()`, and `assertNotHasAnyClass()` — so there's an explicit, PHPUnit-throwing alternative to calling
  `has_class()` mid-chain. The existing `assertNodeHasClass()` and `assertNodeHasAnyClass()` are kept as deprecated aliases for  BC.

In summary, in that PR:
  1. No new logic, no change to has_class() behaviour, no DOM manipulation.
[...]

PR #883, opened 2026-04-29
TECH-DEPT-700: HTML: improve assertions while chaining 

cobianzo:
## Summary

Ticket:  [#700.](https://github.com/alleyinteractive/mantle-framework/issues/700)

The `HTML` class exposes `has_class(): bool` alongside chainable assertion
methods, making it easy to accidentally call `has_class()` mid-chain and
have it silently return `false` without throwing — giving a false sense
that an assertion passed.

This PR adds four new chainable assertion methods to `HTML\Assertions`:

- `assertHasClass( string ...$class )` — asserts the node has ALL given classes
- `assertNotHasClass( string ...$class )` — asserts the node does NOT have all given classes
- `assertHasAnyClass( string ...$class )` — asserts the node has ANY of the given classes
- `assertNotHasAnyClass( string ...$class )` — asserts the node has NONE of the given classes

The existing `assertNodeHasClass()` and `assertNodeHasAnyClass()` are kept
as deprecated aliases for backwards compatibility.

## Test plan

- [x] New tests covering pass and fail cases for all four methods
- [x] Existing deprecated aliases verified to still work
- [x] Full `HtmlTest` suite passes (44 tests, 80 assertions)
- [x] PHPCS clean

### DIFF
--- a/src/mantle/support/html/trait-assertions.php
+++ b/src/mantle/support/html/trait-assertions.php
@@ -22,9 +22,23 @@ trait Assertions {
 	 *
 	 * @param string ...$class Class names.
 	 */
-	public function assertNodeHasClass( string ...$class ): static {
+	public function assertHasClass( string ...$class ): static {
 		Assert::assertTrue( $this->has_class( ...$class ), sprintf(
 			'Failed asserting that the node has class(es): %s',
+			implode( ', ', $class )
+		) );
+
+		return $this;
+	}
+
+	/**
+	 * Assert that the node does not have all of the specified classes.
+	 *
+	 * @param string ...$class Class names.
+	 */
+	public function assertNotHasClass( string ...$class ): static {
+		Assert::assertFalse( $this->has_class( ...$class ), sprintf(
+			'Failed asserting that the node does not have class(es): %s',
 			implode( ', ', $class )
 		) );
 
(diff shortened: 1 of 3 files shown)


src/mantle/support/html/trait-assertions.php
```diff
@@ -22,7 +22,7 @@ trait Assertions {
 	 *
 	 * @param string ...$class Class names.
 	 */
-	public function assertNodeHasClass( string ...$class ): static {
```
dustinpaluch, line 25: I think we should not change method names unless we are bumping the major version.

cobianzo: Got it @dustinpaluch . We can close the PR and just keep it as a guidance. 
Before that: did you notice that the [old methods are still there](https://github.com/alleyinteractive/mantle-framework/pull/883/changes#diff-f6755fc5c5525b321dc7bc3f6a7d979eb94577e98ad3738cf6ef5efd12edfd71R82) -and fully functional- , they just delegate to the new names internally.

If you hadnt seen it, you can give me feedback. If you already knew, I can close the PR until we Sean confirms what option looks better from both suggestions

dustinpaluch: Oh, that's my mistake, I see now that the original methods still exist. Disregard my comment.

srtfisher (approved):

srtfisher:
@copilot fix the linting error

PR #886, opened 2026-05-21
Preserve $_wp_post_type_features between tests

nlemoine:
When a test calls unregister_post_type(), the post type itself is restored from the snapshot on the next test via the preserved $wp_post_types global, but its features (title, editor, etc.) were not. post_type_supports() then returned false for the supposedly-restored post type, breaking downstream code that depends on it (e.g. plugins reading post titles).

Add _wp_post_type_features to GLOBALS_TO_BACKUP so the features global is captured and restored alongside the post type.

Includes a regression test that unregisters the persistent post type in one iteration and asserts post_type_supports() still returns true on the next.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


srtfisher:
@nlemoine thank you for the contribution! I've been out and will be back this week to review and merge this and your other contribution.

nlemoine:
Thanks for your feedback @srtfisher! no rush, review this when you find the time :)

srtfisher (approved):

PR #887, opened 2026-05-28
Add configurable wp-content directory name via WP_CONTENT_DIR_NAME

nlemoine:
## Summary

`Utils` and the rsync installer hard-code `wp-content` when working out whether the code already sits inside a WordPress install, and when resolving the install root (the config file and `ABSPATH`). That breaks on any custom `WP_CONTENT_DIR`. Bedrock is the obvious case, since it renames the content dir to `app`.

This adds a `WP_CONTENT_DIR_NAME` env var (default `wp-content`) plus a `Utils::content_directory_name()` helper, and routes the detection and resolution code through it.

## What changed

- New `Utils::content_directory_name()`: reads `WP_CONTENT_DIR_NAME`, trims slashes, falls back to `wp-content`.
- Replaced the hard-coded literal where it matters:
  - `wordpress-bootstrap.php`: the install-detection branch and both config-file path resolutions
  - `class-utils.php`: the `ABSPATH` derivation in `setup_configuration()`
  - `trait-rsync-installation.php`: `maybe_rsync_wp_content()` and `is_within_wordpress_install()`
- Kept `wp-content` for the rsync destination and the VIP `vip-config` lookups, since the WordPress installed there always uses the standard layout.

With the var unset nothing changes. `preg_quote('wp-content')` only escapes the hyphen, so the patterns match exactly as before.

## Why, beyond Bedrock

A plugin or theme you develop inside a real install (e.g. `wp-content/mu-plugins/foo`) can't currently use Mantle's isolated install, because `is_within_wordpress_install()` always matches the path. Set `WP_CONTENT_DIR_NAME` to something that isn't in your path and Mantle treats the project as standalone, spinning up its own WordPress instead of climbing into the host. That's the case that led me here.

## Tests

`tests/Testing/UtilsTest.php` covers the default, the env override, and slash normalization. Everything else should behave the same with the var unset.


srtfisher:
@nlemoine believe it's flagging the indention in utils if you can fix it so we can merge this fix. thanks! 

I haven't had the chance to use the testing framework outside of a traditional installation. Perhaps we can make it work out of the box and detect a bedrock setup.

nlemoine:
Thanks for your feedback @srtfisher! I fixed the indention issue. 

> Perhaps we can make it work out of the box and detect a bedrock setup.

I don't think there's an easy/reliable way for this, unless hard-coding a new `app` value. The content directory is something that's only known at runtime (e.g. being able to boot a WP instance), `phpunit` can't reliably be aware of it upfront. Bedrock uses `app` but it could be `anything` since `WP_CONTENT_DIR` constant can be defined by end users.

And it currently solves two issues in one stone:
- ✅ having a content directory that's not standard `wp-content`: `WP_CONTENT_DIR_NAME=app phpunit ...`
- ✅ being able to force installing a fresh WP instance while a package might live inside an existing WP instance and don't want test to pick up that instance: `WP_CONTENT_DIR_NAME=unknown-on-purpose phpunit ...`.

srtfisher (approved):

nlemoine:
Thanks @srtfisher !

Issue #812, opened 2025-10-24
[2.x] Convert codebase to PSR-4


srtfisher:
### Description

Convert the entire (or at least testing) to PSR-4 standards for file names.

### Use Case

As titled.

srtfisher:
Batch 2 up in #894 (stacked on #893): `assets`, `cache`, `container`, `log`, `scheduling`, `types`, `view` are now PSR-4 file-named — 35 files renamed, 4 nested dirs recased to PascalCase, namespaces registered in both root and per-package `composer.json`. Class/trait names unchanged (not a breaking change). Verified: `--strict-psr` clean, runtime resolution confirmed (incl. nested namespaces), `composer lint` green. Remaining: application, console, contracts, database, facade, filesystem, framework, http, http-client, query-monitor, queue, support, testkit.

srtfisher:
Batch 3 — PR #895 migrates **application, console, facade, filesystem, http-client, query-monitor, testkit** to PSR-4 file naming. Class/trait/enum names unchanged (not a breaking change). Verified with `composer dump-autoload --strict-psr`, `composer lint`, and runtime autoloader resolution.

Remaining: contracts (54), database (78), framework (60), http (32), queue (31), support (42) — the heavier core packages.

srtfisher:
Batch 4 — PR for **queue** (31 files) → PSR-4. Includes `Providers/WordPress` nesting. Stacked on #895. Verified with strict-psr dump-autoload + `composer lint`.

srtfisher:
Batch 5 — PR #897 for **http** (32 files) → PSR-4 (incl. `Routing` + Middleware/Events/Concerns nesting). Stacked on #896.

srtfisher:
Batch 6 — PR #898 for **support** (42 files) → PSR-4. `helpers/` function files intentionally left (files-autoload, not PSR-4). Stacked on #897.

srtfisher:
Batch 7 — PR #899 for **contracts** (54 files) → PSR-4. Also unified `Mantle\Contracts\REST_API` namespace casing (one interface used `Rest_Api`; non-breaking, PHP is case-insensitive). Stacked on #898.

srtfisher:
Batch 8 — PR #900 for **framework** (`Mantle\Framework` classes, 50 files) → PSR-4. `App\` scaffolding stubs under resources/application-structure/ intentionally retained + excluded from classmap. Stacked on #899.

srtfisher:
[...]

PR #891, opened 2026-06-09
Convert testing package to PSR-4 file naming

srtfisher:
Part of #812.

## Summary

Migrates `mantle-framework/testing` to be **fully PSR-4 autoloadable**. Files and directories are renamed to match their class/trait names; **class and trait names are unchanged**, so this is **not a breaking change**.

This completes the transition the package's own `autoload.php` documented (*"Mantle 2.0 should be fully PSR-4 compliant, at which point these file-requires will be removed"*). It supersedes the earlier closed attempt in #871, and incorporates the extra touch-points that surfaced there (`rector.php`, `Concerns/Rsync_Installation.php`) plus `phpstan.neon`.

## Changes

- **Directories → PascalCase** (9): `concerns→Concerns`, `attributes→Attributes`, `exceptions→Exceptions`, `expectation→Expectation`, `mail→Mail`, `snapshots→Snapshots`, `doubles→Doubles`, plus nested `Concerns/Snapshots` and `Doubles/Sitemaps`. (`data/` stays — fixtures only, no PHP.)
- **Class/trait files → declared name** (77): e.g. `class-test-case.php → Test_Case.php`, `trait-assertions.php → Concerns/Assertions.php`, `class-spy-rest-server.php → Doubles/Spy_REST_Server.php`.
- **Global shim**: `class-wp-unittestcase.php → wp-unittestcase.php` (declares the global `WP_UnitTestCase`; loaded by explicit `require`, not PSR-4).
- **Removed the manual autoload shim**: deleted the `$autoloaded_classes` require list in `autoload.php` and the `extra.wordpress-autoloader` block from the package `composer.json`. Classes now resolve via Composer's PSR-4 autoloader.
- **Updated internal paths**: `wordpress-bootstrap.php`, `Utils.php`, `Concerns/Rsync_Installation.php`, repo-root `rector.php`, and `phpstan.neon`.

## Verification

- `composer dump-autoload --optimize --strict-psr` — **no PSR-4 violations** in the testing package (only the intentional global `wp-unittestcase.php` shim is skipped, as expected).
- Runtime autoload resolution verified for a representative class/trait from **every** renamed directory (43 symbols) + the package helper functions.
- `phpcs` clean.
- `phpstan` — **no new errors** vs. the base branch (the 3 pre-existing errors are identical on `1.x`).
- Full PHPUnit matrix: deferred to CI (local Vagrant env unavailable). This is a rename-only change with no logic edits.


dlh01 (approved):
🌳

PR #892, opened 2026-06-09
Fix correctness bugs across cache, HTTP, database, queue, console, and support

srtfisher:
## Summary

A focused codebase review surfaced a number of latent correctness bugs across several subsystems. Each change is small, isolated, and addresses a concrete bug (no refactors-for-style). Findings were verified against the actual code paths before fixing.

### Cache (`src/mantle/cache/class-repository.php`)
- `remember()` / `remember_forever()` returned the **bool** result of `put()` on a cache miss instead of the computed value. The WordPress repository does not override these, so this affected the live cache.
- `pull()` overwrote the value with the default before returning, so it **always returned the default** and never deleted/returned the stored value on the base implementation.

### HTTP client (`src/mantle/http-client/class-pending-request.php`)
- `retry()` stored the delay under `options['delay']`, but `send()` reads `options['retry_delay']` — so the configured retry delay was **silently ignored** and retries always ran with no backoff.

### HTTP (`src/mantle/http/…`)
- `Response::morph_to_json()` ignored `Jsonable`/`Arrayable` (handled by `should_be_json()`), `json_encode()`-ing the raw object instead of using `to_json()`/`to_array()`.
- `Request::offsetExists()` did `array + null` → **`TypeError`** when no route parameters were set.
- `Route::get_request_parameters()` called `->all()` on a possibly-null `ParameterBag` (now null-safe, consistent with other call sites).
- `bearer_token()` returned the raw `Authorization` header for non-Bearer tokens instead of `null`.

### Database (`src/mantle/database/…`)
- `Builder::without()` used `array_diff_key` against `array_flip($relations)`, comparing integer keys to relation names — it **removed nothing**. Now diffs by value.
- `get_queued_term_attribute()` returned only `[0]` of the queued terms (copy-pasted from the meta path, which stores `[$value, $update]`); it now returns the full value.
- `Has_One_Or_Many::save()` called `$model->first()` on a single model, firing a spurious unconstrained query (and a possible fatal on an empty table). Now uses `$model::get_object_name()`.
- `Term_Query_Builder::get_query_args()` no longer mutates `$this->limit` from a getter.

### Queue (`src/mantle/queue/providers/wordpress/…`)
- `Provider::pop()` set the job lock to a relative duration (`600`) where an absolute timestamp is expected, so `is_locked()` always returned false. Now `time() + timeout`, matching `Queue_Worker_Job::fire()`.
- `Scheduler::schedule_next_run()` used `max()` instead of `min()`, scheduling up to the concurrency ceiling regardless of how little work was pending. Now caps at `max_concurrent_batches`.

### Console (`src/mantle/console/…`)
- `Application::test()` passed the resolved command **instance** as the `command` input value instead of the name string.
- `option()` used `?:`, discarding falsy option values (`0`, `''`, `false`) in favour of the default. Now uses `??`.

### Support (`src/mantle/support/…`)
- `defer()` had an inverted guard (`if ( $request_sent )`) so it **never** called `fastcgi_finish_request()`/`litespeed_finish_request()` — the response was never flushed before the deferred callback ran.
- `Stringable::split()` defaulted `$limit = 1`, so `preg_split` returned the whole string and **never split**. Now `-1`.
- `Str::mask()` dropped the `$encoding` (and length) argument on the trailing-segment `mb_substr`, corrupting output under non-default encodings.

### Misc
- `Filesystem_Manager::$custom_drivers` is now initialized to `[]`.

## Testing

- `php -l` passes on all changed files.
- Pure-logic transformations (`without()` diff, scheduler `min`, `Str::mask` encoding, `split` default) were verified with standalone scripts.
- The full PHPUnit suite requires a provisioned WordPress + MySQL environment (it installs WP and downloads plugins on bootstrap) and could not be run here. The cache and queue tests that exist do not assert the affected return values, so the bugs were not previously caught.

A few lower-confidence observations (e.g. WordPress cache TTL semantics for past `DateTimeInterface` values, queue retry attempt-counting, closure-command base setup) were intentionally left out as they are design decisions rather than clear bugs.

https://claude.ai/code/session_017q5u1a8e9y1vaiuqkJsXRi

---
_Generated by [Claude Code](https://claude.ai/code/session_017q5u1a8e9y1vaiuqkJsXRi)_


dlh01 (approved):
🌳

PR #900, opened 2026-06-12
Convert framework (Mantle\Framework) to PSR-4

srtfisher:
## Summary

Continues the PSR-4 file-naming migration (#812). Converts the **framework** package's `Mantle\Framework` classes (50 class/trait files).

Files and namespace directories (`Bootstrap/`, `Console/`, `Console/Generators/`, `Events/`, `Exceptions/`, `Http/`, `Manifest/`, `Providers/`) were renamed to match their symbol names. `framework` has no package `composer.json`, so its namespace is registered in the **root** `composer.json` PSR-4 autoload; the root `wordpress-autoloader` catch-all continues to serve the remaining un-migrated packages.

### Scaffolding stubs intentionally untouched

The `App\`-namespaced files under `resources/application-structure/` are **templates published into new applications** (via `vendor:publish`), not autoloaded framework code. They keep their `class-*.php` names (renaming them would change scaffolding output) and are added to `autoload.exclude-from-classmap` so `--strict-psr` does not flag them. The `console/generators/stubs/*.stub` placeholder templates are likewise untouched.

**Class and trait names are unchanged**, so this is not a breaking change.

Updated `phpstan.neon` (`Console/Hook_Usage_Command.php`, `Console/Generators`) and `rector.php` (`Exceptions/Handler.php` skip path) references.

> Stacked on #899 (batch 7). Base retargets to `1.x` as the stack merges.

## Verification

- `composer dump-autoload --optimize --strict-psr` — no new violations; the `resources/` App stubs are correctly excluded.
- `composer lint` (phpcs + phpstan + rector) — all green, exit 0.
- Runtime autoloader resolution confirmed across all nested namespaces.

Part of #812


src/mantle/facade/composer.json
kingkool68, line 7: Do you still need this package since the `extra.wordpress-autoloader` configurations were removed?

kingkool68 (approved):
👍

PR #903, opened 2026-06-18
Complete PSR-4 migration: src/Mantle/ + drop wordpress-autoloader (#812)

### LINKED ISSUE #812: [2.x] Convert codebase to PSR-4

srtfisher:
## Summary

Capstone of #812. Completes the PSR-4 migration by restructuring the source tree and removing the WordPress autoloader entirely.

- **`src/mantle/` → `src/Mantle/`**, with every package directory PascalCased to match its namespace (`http-client` → `Http_Client`, `rest-api` → `REST_API`, `new-relic` → `New_Relic`, `query-monitor` → `Query_Monitor`, `framework-views` → `Framework_Views`, `testing-dependencies` → `Testing_Dependencies`).
- Root autoload collapsed to a single **`"Mantle\\": "src/Mantle/"`** PSR-4 mapping.
- **`alleyinteractive/composer-wordpress-autoloader` removed** from `require` and the `extra.wordpress-autoloader` blocks in the root **and all 29 packages**, plus the `monorepo-builder` `DATA_TO_APPEND`.
- Classes now resolve purely through Composer's native PSR-4 autoloader.

### Not a breaking change

Class/trait/interface/enum names are unchanged, and the **published package names remain lowercase-hyphenated** (e.g. `mantle-framework/http-client`). No consumer edits required.

## Notable details

- **The autoloader plugin is intentionally kept in the root `allow-plugins`.** It is still a *transitive* dependency of `wp-caper`, `logger`, `wp-match-blocks`, and `wp-plugin-loader`, so it stays installed — and a *blocked* Composer plugin aborts `composer install`. It no longer autoloads any Mantle code. (Package-level `allow-plugins` is ignored when a package is consumed as a dependency, and the per-package CI runs `composer install --no-plugins`, so it was removed from the package `composer.json` files.)
- **Split workflows** (`split_monorepo.yml`, `tests.yml`) now build their matrix from the PSR-4 **directory** names (`monorepo-builder packages-json` returns the composer *name*, which no longer matches the dir, and there is no clean name→dir transform for acronyms like `REST_API`). The lowercase published repository name is derived from the directory via `tr '[:upper:]' '[:lower:]' | tr '_' '-'`, so `Http_Client` still publishes to `mantle-framework/http-client`, `Testing_Dependencies` → `mantle-framework/testing-dependencies`, etc.
- `Testing/wp-unittestcase.php` (a global `WP_UnitTestCase` shim loaded via `require_once`, not autoloaded) is added to `exclude-from-classmap`, so `composer dump-autoload --strict-psr` reports **zero** package-level violations.
- CHANGELOG: the per-batch PSR-4 bullets were squashed into a single entry.

## Verification

All run locally and green:

- `composer validate --strict` (root **+ all 29 packages** standalone), `composer monorepo-validate`
- `composer phpcs`, `composer phpstan` (no errors, **no baseline change**), `composer rector` (dry-run)
- `composer dump-autoload --optimize --strict-psr` — zero `src/Mantle` package violations
- Runtime reflection: classes resolve from `src/Mantle/<Pascal>`
- End-to-end split-workflow simulation across all 29 packages: each maps to a valid directory **and** the correct lowercase published repo

The full WordPress PHPUnit matrix runs in CI (this is a rename + autoload-config change).

Closes #812.


dlh01 (approved):
🌳

